PR #12, opened 2022-04-08
No str op

j042:
# Overview
Allow reading spline infos directly from cpp-splines instead of using string ops.  Minor updates  in README & setup.py


j042 (commented):
v

PR #16, opened 2022-06-01
Resizing of control point vector in Python update

jzwar:
# Bugfix
 Splinepy did not check the control point size before updating the spline on the python side. It was all my fault and I beg your forgiveness.


j042:
ummmm, maybe

PR #19, opened 2022-07-25
Splinelib submodule

j042:
# Overview
Removes copy of SplineLib and pugixml and instead, adds them as submodule. pugixml is now part of splinelib's dependency. As a result, submodule udpate requires additional flag `--recursive`.


jzwar (approved):
Looks good, nice work

jzwar:
I only wonder, how the SplineLib needs to be linked, now that we refer to our own fork of it? Isn't that a copyright issue?

PR #23, opened 2022-08-02
add _RequiredProperties

j042:
# Overview
Defines required properties for each spline type. Based on this info, `todict()` and `copy()` does not need to be implemented in child class


splinepy/_spline.py
```diff
@@ -566,56 +671,71 @@ def _check_and_update_c(self,):
...
         # Check if given degree match expected number of degrees
-        if not self.degrees.shape[0] == self.para_dim:
-            raise InputDimensionError(
-                "RDimension mismatch between degrees and parametric dimension"
-            )
+        #if not self.degrees.shape[0] == self.para_dim:
+        #    raise InputDimensionError(
+        #        "RDimension mismatch between degrees and parametric dimension"
+        #    )
+
```
j042, line 696: It seemed a bit redundant, but if this was to fight against possible in-place manipulation, we can keep it.

j042: ++ I will remove comments before merging

jzwar (approved):
Nice work, thanks.
Please remember to delete any remaining comments before merging

PR #27, opened 2022-08-04
Fix check update

j042:
# Overview
`_check_and_update_c()` raised error, due to last minute changes, which uses `control_mesh_resolutions`.  
This PR fixes that.  
Plus, now that we have `required_properties`, `_update_c` and `_update_p` didn't need to be separately implemented. Hence, moved up to `Spline`.


jzwar (approved):
I like how clean it has become. Keeping the updating routines up to date was a real struggle, nice work!

PR #29, opened 2022-08-05
BugFix Order Elevation

jzwar:
Repeated order elevations lead to problems in both bezier types. This was due to the update_p routines which lost track of their pointers. It is fixed now. Tested surfaces and lines with up to 25 consecutive order elevations.


j042 (approved):
thanks! let's hope this is the last fix

PR #30, opened 2022-08-11
Submodules completed

j042:
# Overview
Now, `splinepy`'s dependencies are all listed as submodules. No more meld, no more headaches.

## Summary
- `bezman` as submodule
- `SplineLib` commit update - solves too-deep-tree dilemma

## Outlook
Keep in mind, fully recursive evaluation is still faster for low (~3) degree splines.


j042 (commented):
compiles and passes test

jzwar (approved):

PR #35, opened 2022-08-22
Mini PR to add knot-vector-function

jzwar:
## Added new function to spline base to normalize knot-vector
That's it, there is not much more to say about it

### DIFF
--- a/splinepy/_spline.py
+++ b/splinepy/_spline.py
@@ -23,7 +23,7 @@ class _RequiredProperties:
     Helper class to hold required properties of each spline.
     """
 
-    # name mangling to make direct access very annoying 
+    # name mangling to make direct access very annoying
     __required_spline_properties = {
         "Bezier": ["degrees", "control_points"],
         "RationalBezier": ["degrees", "control_points", "weights"],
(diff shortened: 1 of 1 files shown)


splinepy/_spline.py
```diff
@@ -987,6 +987,25 @@ def remove_knots(self, parametric_dimension, knots, tolerance=1e-8):
...
+
+        Returns
+        -------
+        None
+        """
+        if "knot_vectors" in self.required_properties:
+            for (i, kv) in enumerate(self.knot_vectors):
+                n_kv = [k - kv[0] for k in kv]
+                n_kv = [k / kv[-1] for k in kv]
+                self.knot_vectors[i] = n_kv
```
jzwar, line 1006:
```suggestion
                self._knot_vectors[i] = n_kv
```
Is that better?

splinepy/_spline.py
```diff
@@ -987,6 +987,25 @@ def remove_knots(self, parametric_dimension, knots, tolerance=1e-8):
             )
         )
 
+    def normalize_knot_vectors(self,):
+        """
+        Sets all knotvectors into a range if [0,1], where applicable
```
j042, line 992:
```suggestion
        Sets all knot vectors into a range of [0,1], if applicable
```

j042 (approved):
Must have read my mind. Thanks a lot!

j042 (approved):

PR #39, opened 2022-08-23
minor fix normalize kv and add test

j042:
# Overview
Minor fix of `normalize_knot_vectors()` and adds unittest for it


j042:
at this point, it is safe to say our mistakes

PR #40, opened 2022-08-25
Create test.yml

j042:
# Overview
CI


j042:
thank you @deltaLuki for the initial suggestion!

PR #44, opened 2022-08-26
Reduce Degrees had no return type

jzwar:
`reduced` variable was not returned by `reduce_degrees()` function.


j042 (approved):
nice catch

PR #46, opened 2022-08-30
Format0

j042:
# Overview
new era


jzwar (approved):
Thanks for applying this so quickly.
I have one request: There is no version stated for you clang-format file. Unfortunately, many of the keywords are not transferable to older versions. When I updated to the newest (`16.0.0` - Ubuntu) I got warnings regarding the template syntax

Requested to remove the space before the <> brackets in a template declaration

Issue #47, opened 2022-08-31
Create own logger class for splinepy


danielwolff1:
We would like to have a thin logger functionality for `splinepy`, similar to what has been implemented in [`gustaf`](https://github.com/danielwolff1/daniels-gustaf-fork/blob/main/gustaf/utils/log.py).

danielwolff1:
Since the corresponding PR has been merged, I think this issue can be closed now.

PR #49, opened 2022-08-31
Add thin logger as demanded by #47

danielwolff1:
# Implement thin logging functionality for `splinepy`

This is my first PR, hopefully I'm doing it right...

## Related Issues

* [Create own logger class for `splinepy`](https://github.com/tataratat/splinepy/issues/47)

## Description

* Moved `configure()` from `splinepy.utils` to `splinepy.log`
* Added `debug()`, `info()` and `warning()` methods in `splinepy.log`
* Added a base class `SplinepyBase` implemented in `_base.py`, which is now the base class for all `splinepy` classes and so far only implements logging functions `logd()`, `logi()` and `logw()` utilizing the functionality provided by `splinepy.log` 
* Changed logging in all other classes to use the functions provided by `SplinepyBase` instead of the native `logging` module

### DIFF
--- a/splinepy/_spline.py
+++ b/splinepy/_spline.py
@@ -1,14 +1,11 @@
-import abc
-import logging
 import copy
 import os
-import itertools
 
 import numpy as np
 
 from splinepy import utils
 from splinepy import io
-from splinepy import _splinepy as core
+from splinepy._base import SplinepyBase
 
 
 class InputDimensionError(Exception):
@@ -106,7 +103,7 @@ class _RequiredProperties:
 required_properties = _RequiredProperties()
 
 
-class Spline(abc.ABC):
+class Spline(SplinepyBase):
     """
     Abstract Spline Class.
     """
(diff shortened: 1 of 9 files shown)


splinepy/_spline.py
```diff
@@ -1,14 +1,11 @@
...
-import logging
 import copy
 import os
-import itertools
 
 import numpy as np
 
 from splinepy import utils
 from splinepy import io
-from splinepy import _splinepy as core
```
j042, line 11: What happened to this line? does this work?

splinepy/_spline.py
```diff
@@ -217,7 +214,7 @@ def whatami(self, how_dare_you):
         --------
         None
         """
-        logging.warning("Spline - Excuse me, you cannot tell me what I am.")
+        self.logw("Excuse me, you cannot tell me what I am.")
```
j042, line 217:
```suggestion
        self._logw("Excuse me, you cannot tell me what I am.")
```
missing `_`?

splinepy/_spline.py
danielwolff1: whopsie 🙈

splinepy/_spline.py
danielwolff1: Yeah, my bad, thanks for pointing that out! 👍

j042 (approved):
good stuff

Issue #43, opened 2022-08-26
So many warnings


jzwar:
Since the compiler got a bit more "pedantic", a lot of warnings are thrown during the build of splinepy on the c++ side.

Most of them are relatively unimportant and can be resolved by the compiler, however, it risks overlooking some relavant issues. As is good practice, we should try to get rid of these warnings asap.

j042:
#54 !!!
I will see if I can do something about napf/nanoflann warnings

PR #54, opened 2022-09-01
Get rid of warning

jzwar:
# Minor Code Clean-up

This is to deal with issue #43 .

1. There were many warnings when building splinepy, that made the project look very unprofessional. Most of the warnings were related to comparisons between signed and unsigned types. 
2. There were a class member variables used as counters that were very unspecific, called `i` and `j`, which is problematic as it is unclear what the counters are actually referring to.

- [x] clang-format applied
- [x] unit tests for new feature (not applicable)

### DIFF
--- a/cpp/splinepy/py/py_bspline.hpp
+++ b/cpp/splinepy/py/py_bspline.hpp
@@ -57,10 +57,6 @@ public:
   int para_dim_ = para_dim;
   int dim_ = dim;
 
-  // Counters
-  int i = 0;
-  int j = 0;
-
   // Hr. BSpline himself
   BSpline c_bspline;
   std::shared_ptr<ParameterSpace> c_parameter_space;
@@ -72,25 +68,27 @@ public:
       + ", physical dimension: " + std::to_string(dim);
 
   // Hr. BSpline's python Family
-  py::list p_knot_vectors;
   py::array_t<int> p_degrees;
   py::array_t<double> p_control_points;
+  py::list p_knot_vectors;
 
   // Hr. BSpline's switch to skip_update_c
   bool skip_update = false;
 
-  // Constructor.
+  // Empty constructor
   PyBSpline() {
     py::list empty_list;
     p_knot_vectors.attr("clear")();
     p_knot_vectors.append(empty_list);
   }
+
+  // Constructor
   PyBSpline(py::array_t<int> degrees,
             py::list knot_vectors,
             py::array_t<double> control_points)
       : p_degrees(degrees),
-        p_knot_vectors(knot_vectors),
-        p_control_points(control_points) {
+        p_control_points(control_points),
+        p_knot_vectors(knot_vectors) {
     update_c();
   }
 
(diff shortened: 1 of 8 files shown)


cpp/splinepy/utils/arrays.hpp
```diff
@@ -210,13 +210,14 @@ GaussWithPivot(std::array<std::array<double, para_dim>, para_dim>& A,
 
   // back substitution
   double sum;
-  for (int i{para_dim - 1}; i >= 0; i--) {
+  for (std::size_t i_para_dim{}; i_para_dim < para_dim; i_para_dim++) {
+    std::size_t i{para_dim - 1 - i_para_dim};
```
j042, line 214: what kind of warning was this causing?

j042:
Also, once napf's PR merges, could you update the commit?

cpp/splinepy/utils/arrays.hpp
jzwar: This was causing signed unsigned comparison. In order to make `i_para_dim` an unsigned type, I had to change the order, as unsigned types are always `>=0` making the loop unbreakable. So I just introduced a new variable for indexing

j042: is there warning if you keep the `int`?

jzwar: Yes in a later loop. However, I have a feeling that you are not happy with this implementation. There are two alternatives:
```
for (std::size_t i{para_dim - 1}; i != static_cast<std::size_t>(-1) ; i--)
or
for (std::size_t i{para_dim - 1}; i-- > 0;)
```
Maybe I can convince you of either of these

j042: I guess you just know how to read my mind. I am already convinced. The second one: did you already know that too?

jzwar: Haha :D I kind of regret suggesting the second one, as I find it harder to read because of the first iteration and also it looks really bad. Any chance we can settle on the first one as a compromise?

j042: hey hey, no backsies! :D  
For me, the second one is clear. However the first one, `static_cast<std::size_t>(-1)` confuses me a bit

jzwar: What you mean to say is "you annoyed me so much that I gave in", but i still take it 😆

j042:
just one more thing, could you please keep counter as `i, j, k, l`?

cpp/splinepy/py/py_bspline.hpp
```diff
@@ -57,10 +57,6 @@ class PyBSpline {
   int para_dim_ = para_dim;
   int dim_ = dim;
 
-  // Counters
-  int i = 0;
```
j042, line 61: special thanks x 1000000000 for this one

j042 (approved):
Thank you so much!

PR #64, opened 2022-10-17
Repair and Compiler Warnings - WIP

jzwar:
## Changelog
A couple of things were changed to reduce compiler warnings
- `break`-statements to supress compiler warnings
- adjusted signed and unsigned types 
- deleted some semicolons

However, there were some more things that surfaced when building the new version:

### Fixes
- added missing import of `bezman`'s `SplineGroup`-class as they were used for splitting
- added `pragma once` preprocessor flags in all `hpp` files
- The preprocessor variable `SPLINEPY_MORE` was never set but used in a couple of files, I added it in the `CMakeLists.txt` (this needs checking @j042 )

## Remaining Problems

- [ ]  The build takes an enormous amount of RAM, it ran full twice during multithread-build, which is why I had to limit the number of threads to 1 on my local machine
- [ ] There is still a linker error (I am working on that one)
- [ ] PR for nanoflann for nullptr warning

| :exclamation: Help wanted, if anyone wants to work on this (@deltaLuki @clemens.fricke) |
|:---------------------------|


jzwar:
Soooo, after investigating this for a moment, it seems there is still an issue with this implementation:
- Each `py_<splinetype><id>.cpp` creates a new version of the backend type, which confuses the linker
- Putting all these files into one a single file is a solution, but building it takes so much RAM that it repeatedly crashed my computer)

I would suggest we link the backend types dynamically. Also, the deprecated files should already be deleted in this PR!

I would further suggest, that we include a complete build in the CI (e.g., by extending the matrix)

cpp/splinepy/py/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ else()
             init/py_nurbs9.cpp
             init/py_nurbs10.cpp
     )
+    set(DEFS SPLINEPY_MORE)
```
j042, line 49:
```suggestion
```
this won't be part of #62. In next PR, I will add proper changes to cmakelists.

j042 (commented):
great, thanks! could you please remove `.swp` file?

j042:
This builds if you keep ___minimal___ for current splines and ___more___ for `CoreSpline`, which is sufficient to keep the temporary changes. No need to build full dimensions for current and new. If you'd prefer not have both current and new at the same time, I am okay with adding all python implementations to #62.  
We can manage to keep everything statically linked, so no for dynamically linking.  
Yes for extending CI builds

j042 (approved):

PR #71, opened 2022-11-24
Use only one wf with test token

deltaLuki:
# Description
Update the CI and CD for splinepy. 
Workflow steps: 

1. building splinepy for ubuntu, macOS arm and 86x_64 with the python version 3.8, 3.9, 3.10
2. running test with every version
3. update the docs
4. building wheels for ubuntu, macOS arm and 86x_64 with the python versions 3.8, 3.9 and 3.10. 
5. upload the wheels to PyPI with twine using the `TWINE_TEST_TOKEN`


j042:
we need to keep test workflows

j042:
Thank you, this is a great start.
Can we separate full build test and wheelbuild+deploy in a separate workflow?

j042:
also, we can push to pypi instead of test pypi

PR #75, opened 2022-12-06
CI/CD updates

j042:
# Overview
An alternative for #71, and minor changes pointed out from #72. CI/CD was failing to build wheels for the full versions, presumably because of memory consumption. Here, giant `.cpp` files with many templated classes are split into multiple smaller ones. At the same time, there's an option to explicitly instantiate (most of) templated classes, and it is now a default build option. To disable, use `setup.py --no-explicit`.

## Addressed issues
*  Setup PyPI binary dist
* rm excessive semicolon
* test more build options
* CI/CD fails to build full version


jzwar (approved):
Nice Work @j042

Please run clang-format and itll be fine :ok_hand:

PR #76, opened 2022-12-07
Just some initializations

jzwar:
# Overview
Very tiny PR, only some added braces to not confuse the compiler


j042 (approved):
thank you!

PR #80, opened 2022-12-15
add splinepybase for python

j042:
# Overview
Add base type for splinepy. Enables easy and uniform logging for all inherited classes.

## Showcase 
```python
# multipatch.py

class Multipatch(SplinepyBase):
    ...
    def some_func(self):
        self._logi("this func is for demo")
        self._logd("some_func is called and log is tested")
        self._logw("this func doesn't do much")
```

### DIFF
--- a/splinepy/_base.py
+++ b/splinepy/_base.py
@@ -1,0 +1,17 @@
+"""splinepy/splinepy/_base.py
+
+Base type of splinepy
+"""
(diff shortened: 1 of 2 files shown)


jzwar (approved):
Nice work. LGTM
We can add the Base class successively to all other classes in splinepy

j042:
ok, I will !

PR #84, opened 2022-12-18
Small Bugfix

jzwar:
# CD issues with python version 3.8 and lower
I used a new xml package function to pretty print xml files. This functionality was only added in python version 3.9 and up. I put it into a conditional for this version and hope the issue is fixed with it, I would also be happy if we just omit this line.

Second bugfix was a wrong statement in the spline lib used for derivatives of compositions. 

## Checklists


jzwar:
well, the problem seems to be the order in which attributes are written into the file within the xml object. So:
```
<MultiPatch id="0" parDim="2">
```
becomes
```
<MultiPatch parDim="2" id="0">
```

j042:
current solution (until we find a better one) is to keep xml-indent as an option. for python 3.9+ we compare both indent and no-indent. 3.8, no-indent. We skip gismo-export-test for 3.6 and 3.7.

this PR also expands testing matrix to all python versions we deploy.

j042 (approved):

PR #87, opened 2022-12-22
Bugfix, mixed up the order

jzwar:
# Multipatch bug

Mixed up the order in the connectivity template in the splinepy backend


clemens-fricke (approved):
Seems to be ok.

j042 (changes requested):
I request a test

splinepy/multipatch.py
```diff
@@ -116,6 +116,8 @@ def interfaces(self, con):
             raise ValueError(
                     "Connectivity must be stored in a numpy 2D array."
             )
+        if con.dtype is not np.int32:
+            con = con.astype(np.int32)
```
j042, line 120:
```suggestion
        con = utils.data.enforce_contiguous(con, np.int32)
```

splinepy/utils/data.py
```diff
@@ -218,14 +218,12 @@ def enforce_contiguous(array, dtype=None):
     -------
     contiguous_array: array-like
     """
-    if isinstance(array, np.ndarray):
+    if issubclass(type(array), np.ndarray):
```
j042, line 221: I'd keep is instance

splinepy/utils/data.py
```diff
@@ -218,14 +218,12 @@ def enforce_contiguous(array, dtype=None):
...
         if array.flags["C_CONTIGUOUS"]:
-            return array
-        else:
-            return np.ascontiguousarray(array, dtype=dtype)
+            if array.dtype is dtype:
+                return array
 
-    else:
-        return array
+    return np.ascontiguousarray(array, dtype=dtype)
```
j042, line 226: what are you gonna do is dtype is none?

j042 (changes requested):

splinepy/utils/data.py
j042: check if c contiguous and dtype is None, you return, else return np.ascontiguousarray(array, dtype=dtype). If array is contiguous and dtype is the same, you will just get the same array

splinepy/utils/data.py
jzwar: OK, I just remembered a conversation we had, I wasn't sure about the outcome anymore. Changed it back

j042: thank you!

j042:
I think we have different format setup

j042 (approved):
thanks!

PR #89, opened 2022-12-29
Edit copy

j042:
# Overview
copy() initialize using current __init__() 

## Addressed issues
*  subclasses won't go through proper init

### DIFF
--- a/splinepy/spline.py
+++ b/splinepy/spline.py
@@ -1309,11 +1309,9 @@ class Spline(SplinepyBase, core.CoreSpline):
 
         return dict_spline
 
-    def copy(
-        self,
-    ):
+    def copy(self):
         """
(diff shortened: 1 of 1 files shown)


jzwar (commented):
Can you elaborate on 
`properties_round_trip` please

j042:
> Can you elaborate on
> `properties_round_trip` please

splinepy.Spline gathers properties with setters locally and pass them as kwargs to initialize c++ `PySpline::c_splines_`. By default, properties are copied from initialized c_splines_ to the current splinepy.Spline, thus, `properties_round_trip`. This is there to make properties valid after initializing c_splines_, even if you haven't initialized properties using setters. Consider it as an internal kwarg, with which you can avoid the round trip.

PR #92, opened 2023-01-03
Bugfix Extract Boundary didn't work properly

jzwar:
# Bugfix
should have only worked for boundaries that are extracted from the highest parametric dimension (e.g., dimension 1 in 2D).

Now there is an additional test so it won't happen again - hopefully

## Checklists
* [ ] Added example(s)
* [x] Added test(s)

### DIFF
--- a/tests/test_extract_boundaries.py
+++ b/tests/test_extract_boundaries.py
@@ -1,0 +1,77 @@
+try:
+    from . import common as c
+except BaseException:
+    import common as c
+
+
+class extractBoundariesTest(c.unittest.TestCase):
+    def testExtraction2D(self):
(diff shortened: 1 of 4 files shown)


tests/test_extract_boundaries.py
```diff
@@ -0,0 +1,77 @@
+try:
+    from . import common as c
+except BaseException:
+    import common as c
+
+
+class extractBoundariesTest(c.unittest.TestCase):
+    def testExtraction2D(self):
```
j042, line 8: where did the snake go?

j042 (approved):
great catch! thank you x 423751!

PR #101, opened 2023-02-09
Bugfix for Gismo Export module

jzwar:
# Overview
Load function (never tested before), had a bug. In the same PR, also bugfix and tests for json export.

## Checklists
* [x] Added example(s)
* [x] Added test(s)


jzwar:
Note, that this is just a bugfix. There is still a branch pending that binds more functionality (interfaces, multipatch) to the gismo export module.

splinepy/multipatch.py
```diff
@@ -440,7 +440,7 @@ def boundary_from_function(
         try:
             new_boundary_bools = function(relevant_boundary_centers)
         except BaseException:
-            ValueError(
```
clemens-fricke, line 443: good catch

splinepy/spline.py
```diff
@@ -1319,9 +1319,12 @@ def todict(self, tolist=False):
...
             # prepare list if needed.
-            if isinstance(tmp_prop, np.ndarray) and tolist:
-                tmp_prop = tmp_prop.tolist()  # copies
-                should_copy = False
+            if tolist:
+                if isinstance(tmp_prop, np.ndarray):
+                    tmp_prop = tmp_prop.tolist()  # copies
+                    should_copy = False
+                if p == "knot_vectors":
+                    tmp_prop = [t.tolist() for t in tmp_prop]
```
clemens-fricke, line 1327: What structure is knot vectors before?

clemens-fricke (approved):
Seems to me to be all ok. I believe you have changed the io structure now so that splines are mixed in the list. But since they still know what they are that should be no problem. 

But is the io now still the same? Or do json and gismo now return a list on import and everything else a dict object? Then I would think we should discuss if that should be that way or if (even if it does not really make sense anymore) revert to the dict, to keep a common interface.

The change in spline I do not completely understand why `knot_vector` is handled differently. But I do not know what structure it is originally so I would have to trust you that that makes sense.

Seems so far to be fine if all my remarks non-issues.

splinepy/spline.py
jzwar: it was a list of tracked arrays, but tracked arrays are not serialisable which let to problems with the json module

clemens-fricke: Ok but should we not add the if statement that if the type is tracked array we do this kind of conversion `tolist` to also have any other variables save that might have a tracked array? Maybe also in the future?

clemens-fricke: Can we add a tolist() function to the tracked array?

jzwar: this is a very specific application, so I don't think that this will be necessary. All other variables (ctps, degrees,..) are either list already or are stored as tracked-arrays, which are also interpret-able as numpy ndarrays

jzwar: yes, as it inherits from numpy

clemens-fricke: oh ok now I understand tracked arrays already have that functionality but knot_vector is List[TrackedArray].

Yeah ok that is fine than.

j042, line 1327:
```suggestion
            if tolist and tmp_prop:
                if isinstance(tmp_prop, np.ndarray):
                    tmp_prop = tmp_prop.tolist()  # copies
                    should_copy = False
                if p.startswith("knot_vectors"):
                    tmp_prop = [t.tolist() for t in tmp_prop]
```
in case it is None

jzwar: This will not work, if e.g. tmp_prop is a list-like object, because in most cases it will raise ambiguity-error, I will add a "is not None"

jzwar: In later commits we could also change the default value of the get function to false

PR #104, opened 2023-02-23
Bf typo in function reference and missing tests

jzwar:
# Overview
Basis function derivatives were not properly tested and there was a type in the function derivative reference. Updated tests.

@deltaLuki please check the ci-pipeline, there still seems to be an issue
## Checklists
* [ ] Added example(s)
* [x] Added test(s)


deltaLuki:
#98

deltaLuki:
There is a problem with isort inside pre-commit. A solution is to update the isort to version 5.11.5. [Issue #98 ](https://github.com/tataratat/gustaf/issues/98#issue-1600672263)

jzwar:
If there is no reason not to do it, please create a PR so can have CI up and running again

PR #105, opened 2023-03-01
Clean Up of Jacobian ordering

jzwar:
# Overview
Currently the Jacobian is flipped with what one might have expected.
They are now returned in the form 
```python
import numpy as np
import splinepy as sp
test_spline = sp.Bezier(degrees=[3,3],control_points=np.random.rand(16,3))
query = np.random.rand(1,2)
jacs = test_spline.jacobian(query)
```
Then, the jacobians are stored in the ordering
```math
J_{ijk} = \frac{\partial x_j}{\partial \xi_k}\left(q_{i}\right)
```
, i.e.,
```python
iquery, j_dim, k_paradim = 0,0,0
print(jacs[iquery, j_dim, k_paradim])
```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


tests/test_evaluation.py
```diff
@@ -138,15 +138,94 @@ def test_basis_and_support(self):
...
         # use random query points
         q2D = c.np.random.rand(10, 2)
 
         # Rational Bezier and NURBS are equivalent but use different backends
-        rational_c = self.rational
-        nurbs_c = self.nurbs
+        rational_c = self.rational.copy()
+        nurbs_c = self.nurbs.copy()
         # increase orders for derivatives
         for i in range(2):
```
deltaLuki, line 230: To keep the notation `_` should be used instead of `i` for this loop.

deltaLuki (changes requested):

tests/test_evaluation.py
jzwar: nice catch

deltaLuki (approved):

Issue #102, opened 2023-02-10
CI with warnings


jzwar:
There seems to be some functionality in our CI, that is deprecated. 
Check the warning, e.g., [here](https://github.com/tataratat/splinepy/actions/runs/4141991588)

```
build_and_tests (3.8, ubuntu-20.04)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using 
Environment Files. For more information see: 
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

Would some of the CI expert please check this out :D

jzwar:
Is there any progress on this? This still seems to be an issue, see e.g., [here](https://github.com/tataratat/splinepy/actions/runs/4312993924)

deltaLuki:
GitHub has [introduced a new set of environment files](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) for actions. To fix, we have to change the` setup-python@v3` to `setup-python@v4`

PR #106, opened 2023-03-06
Bug fix action warnings

deltaLuki:
# Overview
Update setup-python version (inside the workflow) from 3 to 4 

## Addressed issues
* #102 

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar (approved):
Thanks for dealing with this. Doesn't look very intrusive to me so I think we are good to go.

PR #108, opened 2023-03-15
New standard in gismo export

jzwar:
# Overview
Wit the new standard introduced [here](https://github.com/gismo/gismo/pull/591) we can assign boundaries directly with the rest of the geometry information. This is very handy if the same mesh is used for different load cases or various boundary conditions. The old standard is still supported, however, the new standard is set to default.

## Showcase
```python
import splinepy as sp
# Define some splines
bsp_el2 = sp.BSpline(
    degrees=[1, 1],
    control_points=[[0, 0], [1, 0], [0, 1], [1, 1]],
    knot_vectors=[[0, 0, 1, 1], [0, 0, 1, 1]],
)
nur_el3 = sp.NURBS(
    degrees=[1, 1],
    control_points=[[1, 0], [2, 0], [1, 1], [2, 1]],
     knot_vectors=[[0, 0, 1, 1], [0, 0, 1, 1]],
)
multipatch_geometry = sp.Multipatch([bsp_el2, nur_el3])
multipatch_geometry.boundaries_from_continuity()
sp.io.gismo.export(
    "test_mesh.xml",                # Filename
    multipatch=multipatch_geometry, # Geometry
    indent=True,                    # Pretty-printing in export
    labeled_boundaries=True,        # boundary with names
)
```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


splinepy/io/gismo.py
```diff
@@ -137,44 +144,63 @@ def export(fname, multipatch=None, indent=True):
...
-            bc.text = "\n".join(
+    if labeled_boundaries:
+        # Export geometric boundaries and set a label
+        for boundary_id, face_id_list in enumerate(
+            multipatch.boundaries, start=1
+        ):
+            boundary_data = ET.SubElement(
+                multipatch_element,
+                "boundary",
+                name="BID" + str(boundary_id),
```
j042, line 155: names are assigned BID# ?
is there way to set an arbitrary name?

jzwar: Not yet, but I can extend thr Multipatch class by a dictionary to specify names.

j042: I see, next PR?

jzwar: Should be a two-liner. I can see what I can do in an hour or so...

j042:
what if `multipatch_geometry.boundaries_from_continuity()` returned computed boundaries?

j042 (approved):

splinepy/io/gismo.py
jzwar: ok, i tried some options, not happy with eiter of those. Next PR please :D

PR #109, opened 2023-03-22
Yet another change to the gismo export

jzwar:
# Extend the export for user-options
There are some model-specific options, like assembler options or boundary conditions. Adding optional parameters to the export, helps avoid modifying the files manually once the simulation is runnng.
 

I am sorry that this is the 5th gismo related PR :

## Showcase
```
import splinepy# Test Output against input
multipatch_geometry = splinepy.Multipatch([some, splines])

# Set some options
gismo_options = [
    {
        "tag": "Rostock",
        "attributes": {"Region": "MP", "population": "unfortunate"},
        "text": "",
        "children": [{"tag": "DM"}],
    }
]

splinepy.io.gismo.export(
  "fileName.xml",
  multipatch=multipatch_geometry,
  indent=False,
  labeled_boundaries=True,
  gismo_options=gismo_options
)

```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


splinepy/io/gismo.py
```diff
@@ -12,6 +12,7 @@ def export(
     multipatch=None,
     indent=True,
     labeled_boundaries=True,
+    gismo_options=None,
```
j042, line 15:
```suggestion
    options=None,
```
seem redundant

splinepy/io/gismo.py
```diff
@@ -305,18 +343,22 @@ def export(
         f.write(file_content)
 
 
-def load(fname):
+def load(fname, get_options=True):
```
j042, line 346:
```suggestion
def load(fname, load_options=True):
```
do you need this? why not just always load options if it is specified?

splinepy/io/gismo.py
```diff
@@ -305,18 +343,22 @@ def export(
...
     Parameters
     ----------
     fname : str
       filename of the gismo xml
+    get_options : bool
+      Retrieve additional options (else - multipatch/geometry only)
 
     Returns
     -------
     spline_dic_list : multipatch
```
j042, line 358:
```suggestion
    multipatch: Multipatch
```

splinepy/io/gismo.py
jzwar: because it requires that a tuple is returned every time. If you are ok with that I can change that

PR #110, opened 2023-03-24
Windows wheels

j042:
# Overview
Modify current implementation and backend implementations to support splinepy for windows.
- Adjust cmake, so that it won't print error message with MSVC
- optimisation flags for MSVC
- compile flags for MSVC
- MSVC requires c++20 -> was having trouble with some constexpr evaluation if this is an issue, take a look at `SplineLib/Sources/Utilities/string_operations.inc l.125`
- edit bezman (a lot of `auto` in return) to satisfy MSVC "grammar"
- edit SplineLib - constexpr, basis function tree access sequence (was computing wrong values), cmake
- adjust IO round trip test - write a file in a `tempfile.TemporaryDirectory`. This is because on windows, you can't access an opened file multiple times. take a look at python [docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile). 

## Addressed issues
* missing windows wheels

## Showcase
1. Get access to a windows machine
2. Log in
3. `pip install splinepy`
5. `python -c "import splinepy"`

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


tests/test_gismo.py
```diff
@@ -824,63 +825,22 @@ def is_top(x):
...
 
-                with open(tmpf.name) as tmp_read:
+                with open(tmpf) as tmp_read:
                     self.assertTrue(
                         c.are_items_same(
                             _gismo_export_ref_2d_indent, tmp_read.readlines()
                         )
                     )
 
-        ########################
```
j042, line 842: Oops

PR #116, opened 2023-04-13
Error print

j042:
# Overview
Error message is now printed at the end of the traceback

## Addressed issue
Error message was printing info before traceback and would only print useless info at the end.


jzwar (approved):
I see you only print the Splinpyerror now. I prefer it that way, too. Aborting the mission was a bit useless :). Thank you for the quick fix.

jzwar:
Please merge this PR after #117 so it's both in the new version

PR #117, opened 2023-04-13
fix ResourceWarning

j042:
# Overview
fixes resource warning from json io. Uses context instead of raw `open()`

## Addressed issues
*  `splinepy/splinepy/io/json.py:30: ResourceWarning: unclosed file <_io.TextIOWrapper name=.....`


jzwar (approved):
I didn't even get this warning. Thanks for making the effort.

PR #122, opened 2023-04-17
add cartesian prod

j042:
# Overview
Adds a function for [cartesian product](https://en.wikipedia.org/wiki/Cartesian_product) as well as a version that performs reversed version of the product. Adapted from a [stackoverflow](https://stackoverflow.com/questions/11144513/cartesian-product-of-x-and-y-array-points-into-single-array-of-2d-points/49445693#49445693)'s discussion.

## Showcase
```python
>>> import numpy as np
>>> import splinepy
>>> a = np.array([0,.5,1])
>>> splinepy.utils.data.cartesian_product((a,a))
array([[0. , 0. ],
       [0.5, 0. ],
       [1. , 0. ],
       [0. , 0.5],
       [0.5, 0.5],
       [1. , 0.5],
       [0. , 1. ],
       [0.5, 1. ],
       [1. , 1. ]])
```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)

### DIFF
--- a/splinepy/utils/data.py
+++ b/splinepy/utils/data.py
@@ -2,6 +2,7 @@
 
 Helps helpee to manage data. Some useful data structures.
 """
+from itertools import accumulate, chain, repeat
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


jzwar (approved):
Thanks for the quick implementation. Looks handy for test case for greville points

PR #124, opened 2023-04-17
Fix cartesian_product

### LINKED ISSUE #123: Add Cartesian is broken
```python
import splinepy as sp
import numpy as np
sp.utils.data.cartesian_product([np.array([0,1,2]),np.array([0,2,4])])
```

Error:
```Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/zwar/Git/gustav_dev/splinepy/splinepy/utils/data.py", line 312, in cartesian_product
    dim_reducing_views[i][..., i] = arrays[i][idx[: n_arr - i]]
ValueError: could not broadcast input array from shape (3,) into shape (2,)

j042:
# Overview
fixes #123 and extends the test. Use 2 ifs. bump version[2]


jzwar (approved):
Thanks for the quick bugfix. The test also looks good with different length arrays

PR #125, opened 2023-04-18
use shared_ptr as holder type

j042:
# Overview
Use [shared_ptr as holder type](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html?highlight=smart%20pointer). This should make some cpp<->python operations simpler. Part of preparations for multi-patch base structure.

## Addressed issues
*  std::vector will be much easier to access than py::list and for that std::shared_ptr can be useful


jzwar (approved):
LGTM!

PR #126, opened 2023-04-18
IGA toolset

jzwar:
# Overview
Provide tools and functions that can help prototype IGA applications. These consist of mappings of basis function derivatives into the physical domain. E.g., with derivative, the spline functions are differentiated with respect to the reference coordinates. In order to get the derivatives with respect to the physical coordinates, the following mappings are applied:
```math
\frac{\partial f}{\partial x_i} = \frac{\partial f}{\partial \xi_j} \frac{\partial \xi_j}{\partial x_i}
```
Second order derivatives are a little more complicated, I will provide technotes in the documentation.

## Addressed issues
*  Features for first release

## Showcase
```
import splinepy as sp
# Define the Geometry
geometry = sp.BSpline(
    degrees=[1, 1],
    control_points=[
        [0.5, 0.0],
        [0.5, 0.5],
        [0.0, 0.5],
        [1.0, 0.0],
        [1.0, 1.0],
        [0.0, 1.0],
    ],
    knot_vectors=[[0, 0, 0.5, 1, 1], [0, 0, 1, 1]],
)

# Define the solution field
solution_field = sp.BSpline(
    degrees=geometry.degrees,
    control_points=np.ones((geometry.control_points.shape[0], 1)),
    knot_vectors=geometry.knot_vectors,
)


# Use refinement
solution_field.elevate_degrees([0, 1, 0, 1])

# Retrieve Hessians
evaluation_points = solution_field.greville_abscissae
mapper = solution_field.geometry_mapper(geometry=geometry)
bf_hessian, support = mapper.basis_hessian_and_support(evaluation_points)
```
Also more efficient functions that compute more than one value at the same time, avoiding duplicated computations.

## Checklists
* [x] Added example(s)
* [ ] Added test(s)


splinepy/spline.py
```diff
@@ -574,6 +574,444 @@ def _get_property(spl, property_):
     return sync_from_core(spl)._data["properties"][property_]
 
 
+class GeometryMapper(SplinepyBase):
```
j042, line 577: let's move this to a separate module? maybe time to bring `helpme` in real life.

j042 (commented):
wow so cool! let's organise modules a bit before we merge

docs/source/conf.py
```diff
@@ -42,7 +43,9 @@
 
 # -- Options for HTML output -------------------------------------------------
 
-html_theme = "furo"
+import pydata_sphinx_theme
+
```
j042, line 47:
```suggestion
```
you don't need this part

jzwar: I already deleted some parts needs force pushing

jzwar:
unit tests still missing, documentation up to date

j042 (commented):
Nice! can you make a separate PR with docs layout changes?

j042:
can you maybe use #130 here? I fear if we don't integrate that now, we probably will never update it

jzwar:
> can you maybe use #130 here? I fear if we don't integrate that now, we probably will never update it

Waiting for #130 to merge.

j042 (approved):
impressive

splinepy/helpme/mapper.py
```diff
@@ -0,0 +1,548 @@
...
+            # the field
+            bf_hessians = np.empty(
+                (
+                    queries.shape[0],
+                    np.prod(self._field_reference.degrees + 1),
+                    self._para_dim,
+                    self._para_dim,
+                )
+            )
+            for i in range(self._para_dim):
```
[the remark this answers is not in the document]
jzwar: yes

PR #127, opened 2023-04-19
separate header for type specific spline functions

j042:
# Overview
Pure re-allocation of codes. another preparation process for multi-patch base


jzwar (approved):
LGTM, although I do not really see a benefit at this point

PR #129, opened 2023-04-20
helpme & multi index 

j042:
# Overview
Adds `MultiIndex` class that helps to find global ids for data that's sequentially listed, but actually representing grid-like structure. In other words, helps find control point ids using numpy's `__getitem__` syntax. name multi_index is motivated by numpy's function [np.ravel_multi_index](https://numpy.org/doc/stable/reference/generated/numpy.ravel_multi_index.html#numpy.ravel_multi_index). This way you don't have to use `%` or `//`.  
Also, as discussed, added `helpme` module. Will be a place to implement functions that helps some uses cases with splines. This means some of the functions / classes implemented in `spline.py` will be moved there in next PR. It won't change any behavior of `Spline` class.

## Addressed issues
*  Easy access to global control point ids (raveled index) wanted

## Showcase
```python
import splinepy

...
boundary0_cps = bspline.cps[bspline.multi_index[0, :]]

# as long as nothing changes with in bspline, you can just get ref
multi_index = bspline.multi_index

# boundary 7 would be 4th dim, first slice
boundary7_cps = bspline.cps[multi_index[:, :, :, 0, ...]]

# since this follow's np's  getitem syntax, you don't need the ellipsis at the end 
boundary7_cps_no_ellipsis = bspline.cps[multi_index[:, :, :, 0]]
assert np.allclose(boundary7_cps, boundary7_cps_no_ellipsis)
```

## Checklists
* [ ] Added example(s) #126 will use this and add it to examples
* [x] Added test(s)

### DIFF
--- a/splinepy/helpme/multi_index.py
+++ b/splinepy/helpme/multi_index.py
@@ -9,6 +9,7 @@ class MultiIndex:
     Performs same task as `np.ravel_multi_index`. However, you can use it with
     __getitem__ like queries.
     """
+
     __slots__ = ("_raveled_indices", "_ndim")
     _ellipsis_type = type(Ellipsis)
 
@@ -22,7 +23,10 @@ class MultiIndex:
         grid_resolutions: array-like
         """
         if not isinstance(grid_resolutions, (tuple, list, np.ndarray)):
-            raise TypeError("grid_resolutions must be array-like type, not" f"{type(grid_resolutions)}")
+            raise TypeError(
+                "grid_resolutions must be array-like type, not"
+                f"{type(grid_resolutions)}"
+            )
 
         # create raveled indices
         raveled = np.arange(np.product(grid_resolutions), dtype="int32")
@@ -31,7 +35,6 @@ class MultiIndex:
         # grid's shape
         self._raveled_indices = raveled.reshape(*grid_resolutions[::-1])
         self._ndim = self._raveled_indices.ndim
-
 
     def __getitem__(self, args):
         """
@@ -70,5 +73,3 @@ class MultiIndex:
                 args.append(s)
 
         return self._raveled_indices.__getitem__(tuple(args[::-1])).ravel()
-
-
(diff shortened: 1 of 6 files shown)


splinepy/helpme/multi_index.py
```diff
@@ -0,0 +1,75 @@
...
+                "grid_resolutions must be array-like type, not"
+                f"{type(grid_resolutions)}"
+            )
+
+        # create raveled indices
+        raveled = np.arange(np.product(grid_resolutions), dtype="int32")
+
+        # to allow general __getitem__ like query, turn them into
+        # grid's shape
+        self._raveled_indices = raveled.reshape(*grid_resolutions[::-1])
```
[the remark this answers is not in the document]
jzwar: Same for read...

splinepy/spline.py
```diff
@@ -1118,6 +1118,33 @@ def coordinate_references(self):
 
         return self._data["coordinate_references"]
 
+    @property
+    def multi_index(self):
+        """
+        Easy control point / weights access using (unraveled) multi index.
```
jzwar, line 1124: That is dangerous, as multi-index changes with simple modifications, such as order-elevation and knot insertion

jzwar (approved):
Nice feature, I will have to change my example files to use this easy layout

splinepy/spline.py
j042: why exactly is it dangerous?

jzwar: If someone accesses `multi_index` then performs order elevation the old multi_index is no longer valid

j042: yes, that's why modification calls deletes all the saved data

splinepy/helpme/multi_index.py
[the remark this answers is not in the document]
j042: try it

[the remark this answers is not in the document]
j042: .T was missing. Will edit

PR #130, opened 2023-04-21
add separate calls for basis, support, basis_deriavtive

j042:
# Overview
In applications, you probably need to evaluate basis quite a few times. However, support doesn't change within the same parametric spaces. This provides a separate call.

## Addressed issues
*  We always get support. Too much support

## Showcase
```python
import splinepy

...

support = spline.support(queries)
basis = spline.basis(queries)
basis_der = spline.basis_derivatives(queries, query_orders)
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/spline.py
+++ b/splinepy/spline.py
@@ -1320,7 +1320,7 @@ class Spline(SplinepyBase, core.CoreSpline):
     @_new_core_if_modified
     def basis(self, queries, nthreads=None):
         """
-        Returns basis function values of given queries.
+        Returns basis function values on the supports of given queries.
 
         Parameters
         -----------
@@ -1366,7 +1366,8 @@ class Spline(SplinepyBase, core.CoreSpline):
     @_new_core_if_modified
     def basis_derivative(self, queries, orders, nthreads=None):
         """
-        Returns derivative of basis functions of given queries.
+        Returns derivative of basis functions evaluated on the supports
+        of given queries.
 
         Parameters
         ----------
--- a/splinepy/spline.py
+++ b/splinepy/spline.py
@@ -1296,9 +1296,54 @@ class Spline(SplinepyBase, core.CoreSpline):
         )
 
     @_new_core_if_modified
+    def support(self, queries, nthreads=None):
+        """
+        Returns basis support ids of given queries.
+
+        Parameters
+        -----------
+        queries: (n, para_dim) array-like
+        n_threads: int
+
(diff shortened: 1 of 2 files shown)


splinepy/spline.py
```diff
@@ -1310,19 +1355,45 @@ def basis_and_support(self, queries, nthreads=None):
...
         return super().basis_and_support(
             queries=queries,
             nthreads=_default_if_none(nthreads, settings.NTHREADS),
         )
 
+    @_new_core_if_modified
+    def basis_derivative(self, queries, orders, nthreads=None):
+        """
+        Returns derivative of basis functions of given queries.
+
```
jzwar, line 1370: Can you add a small comment, that this is just evaluated on the support. Else, one could expect one basis function (derivative) for each ctps. With support in the function name that was clearer

jzwar (approved):
LGTM, thanks for adding this. Technically, it is just code refactoring, so I did not check the functionality and trust you on this one

splinepy/spline.py
```diff
@@ -1310,19 +1355,45 @@ def basis_and_support(self, queries, nthreads=None):
...
 
         return super().basis_and_support(
             queries=queries,
             nthreads=_default_if_none(nthreads, settings.NTHREADS),
         )
 
+    @_new_core_if_modified
+    def basis_derivative(self, queries, orders, nthreads=None):
+        """
+        Returns derivative of basis functions of given queries.
```
j042, line 1369:
```suggestion
        Returns derivative of basis functions evaluated on the supports
        of given queries.
```

splinepy/spline.py
```diff
@@ -1295,10 +1295,55 @@ def jacobian(self, queries, nthreads=None):
...
+
+        return super().support(
+            queries=queries,
+            nthreads=_default_if_none(nthreads, settings.NTHREADS),
+        )
+
+    @_new_core_if_modified
+    def basis(self, queries, nthreads=None):
+        """
+        Returns basis function values of given queries.
```
j042, line 1323:
```suggestion
        Returns basis function values on the supports of given queries.
```

j042 (commented):
like that?

splinepy/spline.py
jzwar: yes! perfect

j042:
doc builds, so here it goes

PR #137, opened 2023-04-25
Uniform return types in IO

jzwar:
# Overview
Some of the io routines returned dictionaries of spline data and others returned splines. Now no longer.

Also - save input for irit files

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/io/ioutils.py
+++ b/splinepy/io/ioutils.py
@@ -97,3 +97,50 @@ def abs_fname(fname):
         fname = os.path.abspath(fname)
 
     return fname
+
+
+def strip_tabs(fname, overwrite=True, tab_expand=2):
+    """Replaces tabs in a text file with spaces
+
+    Parameters
+    ----------
+    fname : string
+      Filename
+    overwrite : bool
+      Inplace modification, otherwise new file with prefix copy is created
+    tab_expand : int
+      Determines how many spaces to be set for each tab
+
+    Returns
+    -------
+    None
+    """
+    if overwrite:
+        out_name = fname
+    else:
+        out_name = "copy." + fname
+    with open(fname) as inputFile:
+        file_contents = inputFile.read()
+        file_contents.replace("\t", " " * tab_expand)
+    with open(out_name, "w") as exportFile:
+        exportFile.write(file_contents)
+
+
+def dict_to_spline(spline_dictionary):
+    """Create a list of splines from a list of dictionaries
+
+    Parameters
+    ----------
+    spline_dictionary : list
+      List of dictionaries
+
+    Returns
+    -------
+    spline_lits : list
+      List of splines in called format (NAME_TO_TYPE)
+    """
+    from splinepy.settings import NAME_TO_TYPE
+    from splinepy.spline import Spline
+
+    spline_list = [Spline(**spd) for spd in spline_dictionary]
+    return [NAME_TO_TYPE[spd.name](spline=spd) for spd in spline_list]
--- a/splinepy/io/ioutils.py
+++ b/splinepy/io/ioutils.py
@@ -99,7 +99,7 @@ def abs_fname(fname):
     return fname
 
 
-def strip_tabs(fname, overwrite=True, tab_expand=2):
+def expand_tabs(fname, overwrite=True, tab_expand=2):
     """Replaces tabs in a text file with spaces
 
     Parameters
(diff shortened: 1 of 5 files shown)


splinepy/io/ioutils.py
```diff
@@ -97,3 +97,50 @@ def abs_fname(fname):
...
+    -------
+    None
+    """
+    if overwrite:
+        out_name = fname
+    else:
+        out_name = "copy." + fname
+    with open(fname) as inputFile:
+        file_contents = inputFile.read()
+        file_contents.replace("\t", " " * tab_expand)
```
clemens-fricke, line 124:
```suggestion
    file_contents.replace("\t", " " * tab_expand)
```

clemens-fricke: Can also be here. Does not need to be in the file context. But can also stay inside.

splinepy/io/irit.py
```diff
@@ -15,8 +16,11 @@ def load(fname):
     Returns
```
clemens-fricke, line 16:
```suggestion
        Path to the irit file to read in.
    save_replace: bool
        Replace all tabs in the irit file. Defaults to True.

    Returns
```

clemens-fricke (approved):
Very nice. Seems to be alright. The first comment is insignificant, but the second is more important for the correct documentation. Please at least add the new parameter to the function.

splinepy/io/ioutils.py
```diff
@@ -97,3 +97,50 @@ def abs_fname(fname):
...
+    Returns
+    -------
+    None
+    """
+    if overwrite:
+        out_name = fname
+    else:
+        out_name = "copy." + fname
+    with open(fname) as inputFile:
+        file_contents = inputFile.read()
```
j042, line 123: `read()` may take too big memory if it tries to read a giant file. loop instead?

splinepy/io/ioutils.py
```diff
@@ -97,3 +97,50 @@ def abs_fname(fname):
...
+      Determines how many spaces to be set for each tab
+
+    Returns
+    -------
+    None
+    """
+    if overwrite:
+        out_name = fname
+    else:
+        out_name = "copy." + fname
```
j042, line 121:
```suggestion
        out_name = "copy_" + fname
```

splinepy/io/ioutils.py
```diff
@@ -97,3 +97,50 @@ def abs_fname(fname):
...
+
+    Returns
+    -------
+    spline_lits : list
+      List of splines in called format (NAME_TO_TYPE)
+    """
+    from splinepy.settings import NAME_TO_TYPE
+    from splinepy.spline import Spline
+
+    spline_list = [Spline(**spd) for spd in spline_dictionary]
```
j042, line 145: wow nice use of `Spline`

splinepy/io/ioutils.py
clemens-fricke: Looping is difficult if the file is written back to the same file.

The only thing one could do is to do the loop and write it to the copy_fname file and depending on whether or not the file is to be overwritten move the copy_file to the original file location. Is there a possibility that these irit files get so big that this might be a problem?

splinepy/io/ioutils.py
jzwar: Thought about deleting this line, as fname could also be a global path

splinepy/io/ioutils.py
jzwar: :hugs:

splinepy/io/ioutils.py
clemens-fricke: After looking at the code in `SplineLib/Sources/InputOutput/irit.cpp`, they also read in the whole file at once. So even if we do some workaround for not reading the entire file completely, the problem persists in the `irit` importer.

splinepy/io/ioutils.py
j042: good point

jzwar: Used @clemens-fricke second suggestion

splinepy/io/ioutils.py
j042: sounds like a `won't fix`

splinepy/io/irit.py
```diff
@@ -2,21 +2,29 @@
...
+def load(fname, save_replace=True):
     """
     Read spline in `.itd` form.
 
     Parameters
     -----------
     fname: str
 
+      Path to the irit file to read in.
+    save_replace: bool
```
jzwar, line 17:
```suggestion
    expand_tabs: bool
```

j042 (approved):

PR #140, opened 2023-05-03
[Draft] [Discussion] Sample List

jzwar:
# Overview
This PR is first meant to discuss the outline of the function sample for multipatch objects. @j042 recently introduced `SplineList` as a way to access splinepy-splines in list form from c++ directly. I would like to provide a way to sample these efficiently, maybe even provide a connectivity for multipatch objects. This is also useful for Multipatch objects.

Questions:
1. What is a good outline for the sample function
    - Do we allow for non-uniform splines, i.e. surfaces and volumes in the same spline-list
    - Do we introduce checks here?
    - Do we sample all splines in one thread and multithread the individual samplings, or do we split up the splines along multiple threads
2. What other functions does Splinelist need?
3. Do we want to provide one mesh for one multipatch if we create a plottable object? / FEM-mesh?

Please let me know what you think

### DIFF
--- a/cpp/splinepy/py/py_spline.hpp
+++ b/cpp/splinepy/py/py_spline.hpp
@@ -1011,7 +1011,7 @@ inline void add_spline_pyclass(py::module& m) {
 
             py::kwargs properties = t[0].cast<py::kwargs>(); // init
             PySpline spl(properties);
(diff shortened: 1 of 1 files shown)


j042:
could you move this to discussion?

jzwar:
#141

PR #142, opened 2023-05-05
use RemoveOneParametricDimension() for slicing

j042:
# Overview
Creating boundary spline can be expensive for BSpline Families, as it will recreate the basis functions. This PR avoids it by just taking pointers to existing data structures (they are already shared_ptrs) for boundary spline. Much cheaper. Will possibly use boundaries to plot in gustaf rather than organized queries.

## Addressed issues
*  Maybe expensive to extract boundaries.

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/cpp/splinepy/splines/helpers/extract.hpp
+++ b/cpp/splinepy/splines/helpers/extract.hpp
@@ -43,23 +43,13 @@ ExtractControlMeshSliceFromIDs(const SplineType& spline,
       using KnotVectors = typename PSpace::KnotVectors_;
       using KnotVector = typename KnotVectors::value_type::element_type;
       using VSpace = typename SelfBoundary::PhysicalSpace_;
-      // PSpace - let it create a new basis. Otherwise, it will make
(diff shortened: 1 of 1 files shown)


jzwar (approved):
Looks good, I checked out the SplineLib changes, besides the usual `int` vs `size_t` debate, it checks out. Thanks for the work. 
Is this frequently used or why is there a need to adapt this?

jzwar:
> Looks good, I checked out the SplineLib changes, besides the usual `int` vs `size_t` debate, it checks out. Thanks for the work. Is this frequently used or why is there a need to adapt this?

Is there a test for sub-dim extraction besides boundary extraction? If not, would be nice to add one.

j042:
1) Or you mean `int` vs `NamedType`? for plotting.
2) no, but there's no sub-dim extraction besides boundary extraction

jzwar:
Can be merged

PR #143, opened 2023-05-10
move boundary center eval to helpers

j042:
# Overview
Moves implementation of `EvalauteBoundaryCenters()` to `helpers/scalar_type_wrappers.hpp`, so we can also directly use that in `SplineList` based queries.


cpp/splinepy/splines/helpers/scalar_type_wrapper.hpp
```diff
@@ -31,6 +31,47 @@ void ScalarTypeEvaluate(const SplineType& spline,
...
+template<typename SplineType, typename OutputType>
+void ScalarTypeEvaluateBoundaryCenters(const SplineType& spline,
+                                       OutputType* output) {
+
+  // Prepare inputs
+  const int para_dim = spline.SplinepyParaDim();
+  const int dim = spline.SplinepyDim();
+  const int n_faces = 2 * para_dim;
+
+  double* queries = new double[n_faces * para_dim];
```
jzwar, line 45: Please don't do this to me.

```suggestion
  std::vector<double> queries(n_faces * para_dim);
```
If you fear performance issues due to assignments, please use reserve and push_back instead <3

cpp/splinepy/splines/helpers/scalar_type_wrapper.hpp
```diff
@@ -31,6 +31,47 @@ void ScalarTypeEvaluate(const SplineType& spline,
...
+        const auto q = .5 * (bounds[j] + bounds[j + para_dim]);
+        queries[2 * i * para_dim + j] = q;
+        queries[(2 * i + 1) * para_dim + j] = q;
+      }
+    }
+  }
+
+  // Evaluate
+  for (int i{}; i < n_faces; ++i) {
+    spline.SplinepyEvaluate(&queries[i * para_dim], &output[i * dim]);
```
jzwar, line 68: Then `data()[]` here0

cpp/splinepy/splines/helpers/scalar_type_wrapper.hpp
```diff
@@ -31,6 +31,47 @@ void ScalarTypeEvaluate(const SplineType& spline,
...
+  // Prepare inputs
+  const int para_dim = spline.SplinepyParaDim();
+  const int dim = spline.SplinepyDim();
+  const int n_faces = 2 * para_dim;
+
+  double* queries = new double[n_faces * para_dim];
+
+  // get parametric bounds
+  // They are given back in the order [min_0, min_1,...,max_0, max_1...]
+  double* bounds = new double[2 * para_dim];
```
jzwar, line 49: Same for bounds

jzwar (approved):
LGTM. I don't think that multithreading makes sense here yet as it is just 6 evaluations. But this might come in handy if we want to multithread this operation for multipatch systems.

cpp/splinepy/splines/helpers/scalar_type_wrapper.hpp
j042: I see that you've found the hidden beauty. what about unique_ptr?

j042: before everything goes out of hand, let's just go with std::vector

PR #144, opened 2023-05-10
add uff

j042:
# Overview
Adds uffpy. Will use its cpp header in backend. At the same time, adds the `uffpy.uffpy()` binding as `splinepy.splinepy_core.uffpy()`

```
git submodule update --init
```


j042 (commented):
anything to add here? I'd merge if this passes tests

j042:
@jzwar

PR #146, opened 2023-05-16
Add sparse matrix to iga examples

deltaLuki:
# Overview
In order to reduce storage and increase performance, a sparse matrix from scipy is used in the iga examples instead of numpy.matrix.

## Addressed issues
--

## Showcase
```
laplacian = scipy.sparse.csc_array((bf_laplacian.flatten(), (row_ids, col_ids)), shape=(n_dofs, n_dofs))

solution_field.control_points = scipy.sparse.linalg.spsolve(-laplacian, rhs).reshape(-1, 1)

```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/examples/iga/collocation_laplace_problem.py
+++ b/examples/iga/collocation_laplace_problem.py
@@ -9,7 +9,8 @@ The refinement will only be applied to the solution field, to make the
 calculations more efficient
 """
 import numpy as np
-
+from scipy.sparse import csc_matrix
+import scipy
 import splinepy as sp
 
 try:
@@ -68,7 +69,11 @@ bf_laplacian, support = mapper.basis_laplacian_and_support(evaluation_points)
 laplacian = np.zeros(
     (evaluation_points.shape[0], solution_field.control_points.shape[0])
 )
-np.put_along_axis(laplacian, support, bf_laplacian, axis=1)
+
+n_dofs = evaluation_points.shape[0]
+row_ids = np.arange(support.shape[0]).reshape(-1,1).repeat(support.shape[1], axis=1).flatten()
+col_ids = support.flatten()
+laplacian = scipy.sparse.csc_array((bf_laplacian.flatten(), (row_ids, col_ids)), shape=(n_dofs, n_dofs))
 
 # Evaluate RHS function
 physical_points = geometry.evaluate(evaluation_points)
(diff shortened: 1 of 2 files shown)


examples/iga/galerkin_laplace_problem.py
```diff
@@ -85,7 +86,7 @@ def map_positions(positions, x_min, x_max, y_min, y_max):
 # Assemble individual elements
 ukv = solution_field.unique_knots
 n_dofs = solution_field.control_points.shape[0]
-system_matrix = np.zeros((n_dofs, n_dofs))
+system_matrix = lil_matrix((n_dofs, n_dofs))
```
jzwar, line 89: > Intended Usage
>LIL is a convenient format for constructing sparse matrices
>
>once a matrix has been constructed, convert to CSR or CSC format for fast arithmetic and matrix vector operations
>
>consider using the COO format when constructing large matrices

There is no conversion

deltaLuki: By putting the system matrix together in a for loop, I think the lil_matrix would be better suited for that.
The problem with the coo_matrix is that it doesn't support slicing directly:

> Disadvantages of the COO format
> does not directly support:
- > arithmetic operations
- > slicing
> 
If we want to use coo_matrix to improve performance, the code gets a bit more complex.

jzwar: I don't say we need to use coo_sparse matrices right away. I am suggesting, that we convert the matrix once the assembly is done in order to use it more efficiently in matrix arithmetics as required for solving

jzwar:
If there is no update to this branch I will close the PR soon

examples/iga/collocation_laplace_problem.py
```diff
@@ -68,7 +70,18 @@ def source_function(x):
...
 )
-np.put_along_axis(laplacian, support, bf_laplacian, axis=1)
+
+n_dofs = evaluation_points.shape[0]
+row_ids = (
+    np.arange(support.shape[0])
+    .reshape(-1, 1)
+    .repeat(support.shape[1], axis=1)
+    .flatten()
+)
```
jzwar, line 80:
```suggestion
row_ids = (
    np.arange(support.shape[0])
    .repeat(support.shape[1], axis=0)
    .flatten()
)
```

PR #149, opened 2023-05-22
Ft list compose

### LINKED ISSUE #148: Compose wording
I've realized, that there's some inconsistency.  

core func:
```cpp
std::shared_ptr<PySpline>
Compose(const std::shared_ptr<PySpline>& inner,
        const std::shared_ptr<PySpline>& outer) { ... }
```
python func:
```python
composed = splinepy_core.compose(self, inner_function)
```
For compose, I guess you'd "compose" `inner_function` into the `outer_function`. This means, core func has the flipped arg naming -- is it correct @jzwar?

### LINKED ISSUE #152: composition_derivative for RationalBezier Fails
wrong return type casting in `splines/rational_bezier.hpp`

j042:
# Overview
* Adds `compose` and `composition_derivative` to list functions.
* `cartesian_product` option is added to avoid nested for loop. 
* Adds `NThreadQueryType` option to `NThreadExecution` to support "step" execution 
* fixes #148
* fixes #152

WIP - bindings and tests .......

## Addressed issues
* slow to loop in python

## Showcase
```python
import splinepy
from splinepy.splinepy_core import lists

outer_splines = [<list-of-splines0>]
inner_splines = [<list-of-splines1>]
composed = lists.compose(outer_splines, inner_splines, cartesian_product=True, nthreads=300)

assert len(composed) = len(outer_splines) * len(inner_splines)
```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)

### DIFF
--- a/cpp/splinepy/py/py_spline_extensions.hpp
+++ b/cpp/splinepy/py/py_spline_extensions.hpp
@@ -86,11 +86,11 @@ inline std::shared_ptr<PySpline> Add(const std::shared_ptr<PySpline>& a,
 
 /// spline composition - currently only for bezier
 inline std::shared_ptr<PySpline>
-Compose(const std::shared_ptr<PySpline>& inner,
-        const std::shared_ptr<PySpline>& outer) {
+Compose(const std::shared_ptr<PySpline>& outer,
+        const std::shared_ptr<PySpline>& inner) {
   // performs runtime checks and throws error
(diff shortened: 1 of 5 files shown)


j042:
@jzwar

j042:
these implementations are less relevant in following prs, so i'll just merge now

PR #150, opened 2023-05-23
WIP: Extend spline documentation

danielwolff1:
# Overview
Started to adapt the spline documentation. 

## Addressed issues
*  #132 

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/bezier.py
+++ b/splinepy/bezier.py
@@ -4,6 +4,57 @@ from splinepy import settings, spline, splinepy_core
 class BezierBase(spline.Spline):
     """Bezier Base. Contain extra operations that's only
     available for bezier families.
+
+    Bezier splines are a special type of spline where the basis functions have
+    a global support. The basis functions are given by the so-called
+    *Berstein polynomials*
+
+    .. math::
+            N_{i;p}(u) = \\binom{p}{i} u^i (1-u)^{p-i}
+
+    A Bezier spline itself can be seen as a map from the parametric domain 
+    :math:`\\[0,1\\]^{N_{param}}` to the physical domain 
+    :math:`\\mathbb{R}^{N_{phys}}`. We can distinguish between different types 
+    of splines depending on the dimension of the parametric space.
+
+    #. A spline of degree :math:`p` with control points
+    :math:`P_i\\in\\mathbb{R}^{N_{phys}` and a one-dimensional parameter space
+    corresponds to a line embedded into the physical space:
+
+    .. math::
+            C(u) = \\sum_{i=0}^{l} N_{i;p}(u) P_i
+
+    #. A spline of degrees :math:`p,q` with control points
+    :math:`P_{i,j}\\in\\mathbb{R}^{N_{phys}` and a two-dimensional parameter
+    space corresponds to a surface, embedded into the physical space:
+
+    .. math::
+            C(u,v) = \\sum_{i=0}^{l} \\sum_{j=0}^{m} N_{i;p}(u) N_{j;q}(v) 
+                P_{i,j}
+
+    Due to the tensor-product nature of the Bezier basis functions, this is
+    often rewritten as in terms of multi-variate basis functions
+
+    .. math::
+            \\tilde{N}_{i,j;p,q}(u,v) \\coloneqq N_{i;p}(u) N_{j;q}(v)
+
+    #. A spline of degrees :math:`p,q,r` with control points
+    :math:`P_{i,j,k}\\in\\mathbb{R}^{N_{phys}` and a three-dimensional
+    parameter space corresponds to a surface, embedded into the physical space:
+
+    .. math::
+            C(u,v,w) = \\sum_{i=0}^{l} \\sum_{j=0}^{m} \\sum_{k=0}^{n} 
+                N_{i;p}(u) N_{j;q}(v) N_{k;r}(w) P_{i,j,k}
+
+    Here, we can introduce the multi-variate basis functions
+
+    .. math::
+            \\tilde{N}_{i,j,k;p,q,r}(u,v,w) \\coloneqq N_{i;p}(u) N_{j;q}(v) 
+                N_{k;r}(w)
+    
+    Finally, we like to emphasie that Bezier splines can be seen as a special 
+    type of B-Splines, where the knot vectors do not feature internal knots 
+    but only have the first and last entry :math:`p^i+1`-times repeated.
     """
 
     __slots__ = ()
@@ -170,6 +221,9 @@ class Bezier(BezierBase):
     def __init__(self, degrees=None, control_points=None, spline=None):
         """
         Bezier (Spline).
+
+        See :class:`.BezierBase` for more information on the theory of Bezier
+        splines.
 
         Parameters
         -----------
--- a/splinepy/bezier.py
+++ b/splinepy/bezier.py
@@ -17,7 +17,7 @@ class BezierBase(spline.Spline):
     We can distinguish between different types of splines depending on the
     dimension of the parametric space.
 
-    1. A spline of degree :math:`p` with control points
+    1. A spline of degree :math:`p` with :math:`(p+1)` control points
     :math:`P_i\\in\\mathbb{R}^{N_{phys}}` and a one-dimensional parameter space
     (i.e., :math:`N_{param}=1`) corresponds to a line embedded into the
     physical space:
@@ -59,6 +59,33 @@ class BezierBase(spline.Spline):
     special type of B-Splines with open knot vectors (i.e., the first and last
     entry are repeated :math:`p+1`-times) that do not feature additional
     internal knots.
+    
+    Usage:
+    
+    .. code-block:: python
+       
+       # Polynomial Bezier surface
+       polynomial_bezier = splinepy.Bezier(
+           degrees=[2,1],
+           control_points=[
+               [0.0, 0.0],
+               [1.0, 0.0],
+               [2.0, 1.0],
+               [0.0, 2.0],
+               [1.0, 1.0],
+               [2.0, 2.0]
+           ]
+       )
+       # Rational bezier surface (quarter arc)
+       rational_bezier = splinepy.RationalBezier(
+           degrees=[2],
+           control_points=[
+               [0.0, 0.0],
+               [1.0, 0.0],
+               [1.0, 1.0]
+           ],
+           weights=[1.0, 2.0**-0.5, 1.0]
+       )
     """
 
     __slots__ = ()
--- a/splinepy/bezier.py
+++ b/splinepy/bezier.py
@@ -10,51 +10,55 @@ class BezierBase(spline.Spline):
     *Berstein polynomials*
 
     .. math::
-            N_{i;p}(u) = \\binom{p}{i} u^i (1-u)^{p-i}
+            B_{i;p}(u) = \\binom{p}{i} u^i (1-u)^{p-i}
 
-    A Bezier spline itself can be seen as a map from the parametric domain
-    :math:`[0,1]^{N_{param}}` to the physical domain
-    :math:`\\mathbb{R}^{N_{phys}}`. We can distinguish between different types
-    of splines depending on the dimension of the parametric space.
+    The parametric domain of a Bezier spline is always a multi-dimensional 
+    hypercube, that is :math:`\Omega_{param}=[0,1]^{N_{param}}`. 
+    We can distinguish between different types of splines depending on the 
+    dimension of the parametric space.
 
-    #. A spline of degree :math:`p` with control points
-    :math:`P_i\\in\\mathbb{R}^{N_{phys}}` and a one-dimensional parameter space
-    corresponds to a line embedded into the physical space:
-
-    .. math::
-            C(u) = \\sum_{i=0}^{l} N_{i;p}(u) P_i
-
-    #. A spline of degrees :math:`p,q` with control points
-    :math:`P_{i,j}\\in\\mathbb{R}^{N_{phys}}` and a two-dimensional parameter
-    space corresponds to a surface, embedded into the physical space:
+    1. A spline of degree :math:`p` with control points
+    :math:`P_i\in\mathbb{R}^{N_{phys}}` and a one-dimensional parameter space
+    (i.e., :math:`N_{param}=1`) corresponds to a line embedded into the 
+    physical space:
 
     .. math::
-            C(u,v) = \\sum_{i=0}^{l} \\sum_{j=0}^{m} N_{i;p}(u) N_{j;q}(v)
+            S^B(u) = \sum_{i=0}^{l} B_{i;p}(u) P_i
+
+    2. A spline of degrees :math:`p,q` with control points
+    :math:`P_{i,j}\in\mathbb{R}^{N_{phys}}` and a two-dimensional parameter
+    space (i.e., :math:`N_{param}=2`) corresponds to a surface, embedded into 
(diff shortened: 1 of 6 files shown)


danielwolff1:
This is WIP for now but since I did not get feedback if the amount of documentation is too detailed or too superficial, I wanted to prepare this merge request so that you have easier comparison, I'll keep working on that.

splinepy/bezier.py
```diff
@@ -4,6 +4,57 @@
 class BezierBase(spline.Spline):
     """Bezier Base. Contain extra operations that's only
     available for bezier families.
+
+    Bezier splines are a special type of spline where the basis functions have
```
jzwar, line 8:
```suggestion
    Beziers are a special type of spline where the basis functions have
```

splinepy/bezier.py
```diff
@@ -4,6 +4,57 @@
...
+    a global support. The basis functions are given by the so-called
+    *Berstein polynomials*
+
+    .. math::
+            N_{i;p}(u) = \\binom{p}{i} u^i (1-u)^{p-i}
+
+    A Bezier spline itself can be seen as a map from the parametric domain
+    :math:`\\[0,1\\]^{N_{param}}` to the physical domain
```
jzwar, line 16:
```suggestion
    :math:`[0,1]^{N_{param}}` to the physical domain
```
Original did not work on my machine, please confirm that this is still valid on mac.

splinepy/bezier.py
```diff
@@ -4,6 +4,57 @@
...
+    :math:`\\mathbb{R}^{N_{phys}}`. We can distinguish between different types
+    of splines depending on the dimension of the parametric space.
+
+    #. A spline of degree :math:`p` with control points
+    :math:`P_i\\in\\mathbb{R}^{N_{phys}` and a one-dimensional parameter space
```
jzwar, line 21:
```suggestion
    :math:`P_i\\in\\mathbb{R}^{N_{phys}}` and a one-dimensional parameter space
```

splinepy/bezier.py
```diff
@@ -4,6 +4,57 @@
...
+    corresponds to a line embedded into the physical space:
+
+    .. math::
+            C(u) = \\sum_{i=0}^{l} N_{i;p}(u) P_i
+
+    #. A spline of degrees :math:`p,q` with control points
+    :math:`P_{i,j}\\in\\mathbb{R}^{N_{phys}` and a two-dimensional parameter
```
jzwar, line 28:
```suggestion
    :math:`P_{i,j}\\in\\mathbb{R}^{N_{phys}}` and a two-dimensional parameter
```

splinepy/bezier.py
```diff
@@ -4,6 +4,57 @@
...
+
+    .. math::
+            C(u,v) = \\sum_{i=0}^{l} \\sum_{j=0}^{m} N_{i;p}(u) N_{j;q}(v)
+                P_{i,j}
+
+    Due to the tensor-product nature of the Bezier basis functions, this is
+    often rewritten as in terms of multi-variate basis functions
+
+    .. math::
+            \\tilde{N}_{i,j;p,q}(u,v) \\coloneqq N_{i;p}(u) N_{j;q}(v)
```
jzwar, line 39:
```suggestion
            \\tilde{N}_{i,j;p,q}(u,v) := N_{i;p}(u) N_{j;q}(v)
```
Coloneqq is unknown to mathjax, please use := instead

splinepy/bezier.py
```diff
@@ -4,6 +4,57 @@
...
+
+    #. A spline of degrees :math:`p,q,r` with control points
+    :math:`P_{i,j,k}\\in\\mathbb{R}^{N_{phys}` and a three-dimensional
+    parameter space corresponds to a surface, embedded into the physical space:
```
jzwar, line 43:
```suggestion
    parameter space corresponds to a volume, embedded into the physical space:
```
I think we got the idea with surfaces already, but this can stay. Please also mention, that parametric and physical domain can be different in a general case and that we support up to 10 dimensions for both the parametric and physical domain seperately

jzwar:
image
For reference

danielwolff1:
Thanks for the fast review! 

Since I am now finally able to build the documentation locally, I will fix the issues with the rendering asap!

splinepy/bezier.py
```diff
@@ -171,6 +222,9 @@ def __init__(self, degrees=None, control_points=None, spline=None):
         """
         Bezier (Spline).
 
+        See :class:`.BezierBase` for more information on the theory of Bezier
+        splines.
+
         Parameters
         -----------
         degrees: (para_dim,) list-like
```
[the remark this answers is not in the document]
danielwolff1: Screenshot 2023-05-24 at 10 09 14

Are you sure? For me it displays perfectly fine, or I don't get what you mean... 🤔

splinepy/bezier.py
danielwolff1: In the documentation, they are not assumed to be equal, that's why different symbols for the parametric and physical domains are used. But I'm now going to add additional information to the documentation and store them in the `Spline` class.

splinepy/bezier.py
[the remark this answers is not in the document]
jzwar: There are two blocks of docstring. The one attached to the constructor is not displayed. That's why I think the docstring is supposed to be written into the class header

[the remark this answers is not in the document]
jzwar: Check out the Bezier class fir reference

splinepy/bezier.py
```diff
@@ -4,6 +4,61 @@
 class BezierBase(spline.Spline):
     """Bezier Base. Contain extra operations that's only
     available for bezier families.
+
+    Beziers are a special type of spline where the basis functions have
+    a global support. The basis functions are given by the so-called
+    *Berstein polynomials*
+
+    .. math::
+            B_{i;p}(u) = \\binom{p}{i} u^i (1-u)^{p-i}
```
jzwar, line 13: You can do without the double backslashes by using raw strings. 
```python 
r"""
Something with \
"""

File: splinepy/bezier.py [hunk 9]
```diff
@@ -4,6 +4,61 @@
 class BezierBase(spline.Spline):
     """Bezier Base. Contain extra operations that's only
     available for bezier families.
+
```
jzwar, line 7: Can you please add the "s" in the line above, I know it was probably my mistake, but I just noticed it :wink:

splinepy/bezier.py
```diff
@@ -4,6 +4,61 @@
...
+    .. math::
+            B_{i;p}(u) = \\binom{p}{i} u^i (1-u)^{p-i}
+
+    The parametric domain of a Bezier spline is always a multi-dimensional
+    hypercube, that is :math:`\\Omega_{param}=[0,1]^{N_{param}}`.
+    We can distinguish between different types of splines depending on the
+    dimension of the parametric space.
+
+    1. A spline of degree :math:`p` with control points
```
jzwar, line 20:
```suggestion
    1. A spline of degree :math:`p` with :math:`(p+1)` control points
```

splinepy/bezier.py
```diff
@@ -4,6 +4,61 @@
...
+    :math:`P_i\\in\\mathbb{R}^{N_{phys}}` and a one-dimensional parameter space
+    (i.e., :math:`N_{param}=1`) corresponds to a line embedded into the
+    physical space:
+
+    .. math::
+            S^B(u) = \\sum_{i=0}^{l} B_{i;p}(u) P_i
```
jzwar, line 26: This is just a minor change, but I suggest to stick to the general naming convention used by GE and also NURBS book, where curves are `C`, Surfaces `S` and Volumes `V`, higher dimensional objects are then written as `M` for multivariate.

I don't mind if you stick to the current notation, just wanted to put it out there.

splinepy/bezier.py
```diff
@@ -165,6 +220,13 @@ def composition_derivative(self, inner, inner_derivative):
 
 
 class Bezier(BezierBase):
+    """
+    Bezier (Spline).
+
+    See :class:`.BezierBase` for more information on the theory of Bezier
+    splines.
+    """
```
jzwar, line 228: Please also move the init docstring here so the parameters are described in the documentation.

splinepy/bezier.py
```diff
@@ -4,6 +4,61 @@
...
+    Here, we can introduce the multi-variate basis functions
+
+    .. math::
+            \\tilde{B}_{i,j,k;p,q,r}(u,v,w) := B_{i;p}(u) B_{j;q}(v)
+                B_{k;r}(w)
+
+    Finally, we like to emphasize that Bezier splines can also be seen as a
+    special type of B-Splines with open knot vectors (i.e., the first and last
+    entry are repeated :math:`p+1`-times) that do not feature additional
+    internal knots.
```
jzwar, line 61: What do you think of adding a simple example to show how it is to be used:
```suggestion
    internal knots.
    
    Usage:
    
    .. code-block:: python
       
       # Polynomial Bezier surface
       polynomial_bezier = splinepy.Bezier(
           degrees=[2,1],
           control_points=[
               [0.0, 0.0],
               [1.0, 0.0],
               [2.0, 1.0],
               [0.0, 2.0],
               [1.0, 1.0],
               [2.0, 2.0]
           ]
       )
       # Rational bezier surface (quarter arc)
       rational_bezier = splinepy.RationalBezier(
           degrees=[2],
           control_points=[
               [0.0, 0.0],
               [1.0, 0.0],
               [1.0, 1.0]
           ],
           weights=[1.0, 2.0**-0.5, 1.0]
       )
```

splinepy/bspline.py
```diff
@@ -4,8 +4,82 @@
...
+    .. math::
+            S^N(u) = \sum_{i=1}^{l} N_{i;p}(u) P_i
+
+    2. A B-Spline of degrees :math:`p,q` with control points
+    :math:`P_{i,j} \in \mathbb{R}^{N_{phys}}` and a two-dimensional parameter
+    space (:math:`N_{param}=2`) corresponds to a surface, embedded into the
+    physical space:
+
+    .. math::
+            S^N(u,v) = \sum_{i=1}^{l} \sum_{j=1}^{m} N_{i;p}(u) N_{j;q}(v)
```
jzwar, line 61: See above

splinepy/nurbs.py
```diff
@@ -2,8 +2,83 @@
...
+    NURBS are an extension of B-Splines overcoming their drawback of being
+    unable to represent circular shapes. This is achieved by the introduction
+    of a weighting function
+    :math:`W^N:\mathbb{R}^{N_{param}}\to\mathbb{R}^{> 0}`, defined as (for
+    a one-dimensional parameter space, i.e., :math:`N_{param}=1`)
+
+    .. math::
+            W^N(u) = \sum_{i=1}^{l} N_{i;p}(u) w_i
+
+    with scalar weights :math:`w_i\in\mathbb{R}^{> 0}`.
```
jzwar, line 17:
```suggestion
    with scalar weights :math:`w_i\in\mathbb{R}^{> 0}`. Note, that the same basis functions are used for both the weighting function and projection space.
```

splinepy/bspline.py
```diff
@@ -4,8 +4,82 @@
...
+            \Theta_u = \{u_1,u_2,...,u_{l+p+1}\}
+
+    such that
+
+    .. math::
+            u_1 \leq u_2 \leq ... \leq u_{l+p+1}
+
+    Here, :math:`l` denotes the number of basis functions/control points along
+    this parametric dimension and :math:`p` the polynomial order of these
+    basis functions.
```
jzwar, line 26: I know it's a bit of an overkill, but can you add the relation of the expected number of control points, knots and the degree respectively

jzwar (commented):
Thank you so much for taking the time to extend the documentation. Just a couple of general remarks:
1. The notation used here differs in some aspects from the one used in mapper, especially when it comes to the upper index notation. I don't really care which notation is used (semi-colon or upper index) so long as it is consistent.
2. Some of the links are missing and it might be good to have a short commentary on what dimensions are supported, just in case someone (ehm - like me) tries to use 64-dimensional Splines.
3. I personally like the idea of including small examples of how the different splines are generated, just so you don't have to jump back and forth to find the appropriate example. This can be a separate PR, in which case please write an issue.

Once these changes are implemented, this can be approved.

splinepy/bezier.py
j042: Probably my mistake. Also "'re"  (`Contains extra operations that're --`)

splinepy/bezier.py
j042: sphinx default is not to document any dunder methods. I once added a workaround to document `__init__` and it seems that I removed it in 646741a. Would you rather just revive that or document all of them in class doctring?

jzwar: I don't think an explicit documentation of the `__init__` and other python specifics makes a lot of sense. I'd rather have the documentation within the class header.

danielwolff1: Okay, then I'll move the `__init__()` methods to the class doc-strings to resolve this.

splinepy/bezier.py
danielwolff1: Alright, I can change that.

splinepy/bezier.py
danielwolff1: Sorry guys, I didn't spot the mistakes while extending the documentation, otherwise I would have changed it for sure, now I'm gonna adapt it after this review :)

splinepy/bezier.py
danielwolff1: Sure, I can do that, I'll copy examples from the examples folder.

splinepy/bspline.py
```diff
@@ -4,8 +4,82 @@
 
 
 class BSplineBase(spline.Spline):
-    """BSpline base. Contains extra operations that's only available for
+    r"""BSpline base. Contains extra operations that are only available for
```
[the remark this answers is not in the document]
danielwolff1: Shouldn't that be clear from the fact that `NURBS` inherits from `BSplineBase`? But sure, I can add that comment

splinepy/nurbs.py
```diff
@@ -2,8 +2,83 @@
...
 
 class NURBS(BSplineBase):
-    """
+    r"""
     Non-Uniform Rational B-Spline.
+
+    NURBS are an extension of B-Splines overcoming their drawback of being
+    unable to represent circular shapes. This is achieved by the introduction
+    of a weighting function
+    :math:`W^N:\mathbb{R}^{N_{param}}\to\mathbb{R}^{> 0}`, defined as (for
```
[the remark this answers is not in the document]
danielwolff1: What index supposed to mean, I haven't seen that before :D

danielwolff1:
Thanks for the extensive review guys. I'll address your points and come back to you.

danielwolff1:
I added examples and addressed (most) points of the review. Next, I'm gonna adopt the notation to the one in `mapper`, i.e., basically, I'm gonna replace all subscripts with superscripts. This is going to look weird, but it's the easiest non-intrusive fix to make the notations compatible. 

(For the theory part it is important to understand which basis function has which degree, while this is not required when writing down the equations for transforming derivatives between the parametric and the physical domain. If I only made the summation index a superscript and let the polynomial degree as a subscript, it would not be consistent because the mapper class uses subscripts only for components).

j042 (approved):
Again, this is awesome. Thank you so much @danielwolff1! I will just add minor format fix and merge

PR #151, opened 2023-05-24
Bezman Composition Sensitivities binding

jzwar:
# Overview
In order to compute the adjoints for microstructures with modifiable deformation function, we need the derivative of a composition with respect to the outer geometries control points.

## Addressed issues
*  Required for Hutchinson TC

## Showcase
```python
import splinepy as sp
import numpy as np

outer = sp.Bezier(
            degrees=[1, 2],
            control_points=[[1, 1], [2, 1], [1, 2], [2, 2], [1, 3], [2, 3]],
        )
inner = sp.RationalBezier(
            degrees=[2, 1],
            control_points=[
                [0.2, 0.0],
                [0.5, 0.5],
                [1.0, 0.2],
                [0.0, 0.8],
                [0.5, 0.5],
                [0.8, 1.0],
            ],
            weights=[1, 0.5, 1, 1, 0.5, 1],
        )

composed_geometry = outer.compose(inner)
list_of_control_point_derivatives = outer.compose_sensitivities(inner)
```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


tests/test_composition_sensitivities.py
```diff
@@ -0,0 +1,76 @@
+try:
+    from . import common as c
+except BaseException:
+    import common as c
+
+
+class compositionSensitivitiesTest(c.unittest.TestCase):
```
j042, line 7:
```suggestion
class CompositionSensitivitiesTest(c.unittest.TestCase):
```

splinepy/bezier.py
```diff
@@ -100,6 +100,87 @@ def compose(self, inner_function):
...
+            self, inner_function
+        )
+
+        return [
+            settings.NAME_TO_TYPE[cc.name](spline=cc)
+            for cc in composed_sensivities
+        ]
+
+    @spline._new_core_if_modified
+    def composition_derivative(self, inner, inner_derivative):
```
j042, line 144: nice doc

cpp/splinepy/splines/bezier.hpp
```diff
@@ -345,6 +345,133 @@ class Bezier : public splinepy::splines::SplinepyBase,
...
+      return to_return;
+    }
+  }
+
+  /// Spline composition.
+  /// inner_function requirements:
+  ///   1. Bezier Types
+  ///   2. dim is same as outer_function's par_dim
+  virtual std::vector<std::shared_ptr<SplinepyBase>>
+  SplinepyComposeSensitivities(
```
j042, line 382: Either both `ComposeSentitivieis` & `ComposeDerivative` or `Composition---`. I am leaning towards the second one

cpp/splinepy/py/py_spline_extensions.hpp
```diff
@@ -66,6 +66,21 @@ Compose(const std::shared_ptr<PySpline>& inner,
...
+/// spline composition sensitivities - currently only for bezier
+inline py::list ComposeSensitivities(const std::shared_ptr<PySpline>& inner,
+                                     const std::shared_ptr<PySpline>& outer) {
+  // performs runtime checks and throws error
+  py::list returnable{};
+
+  // Convert into python readable type
+  const auto temp_shared_ptr_vector =
+      inner->Core()->SplinepyComposeSensitivities(outer->Core());
+  for (size_t i{}; i < temp_shared_ptr_vector.size(); i++) {
```
j042, line 78: possibly replace it with conversion helperfunc from spline list prs

j042 (commented):
I just realized you named the PR `CompositionSensitivities`, so I guess this answers one of the question below

jzwar:
@j042 How would you feel about having a bool flag in the compose function that returns the sensitivities when requested? That way, we don't need another function.

j042:
> @j042 How would you feel about having a bool flag in the compose function that returns the sensitivities when requested? That way, we don't need another function.

sounds good!

j042 (approved):
awesome, thank you !

PR #167, opened 2023-06-14
Updated all reduce/elevate degree calls

jzwar:
# Overview
There were deprecated calls to reduce_degree and elevate_degree, which were only aliases and failed sometimes.


j042 (approved):
nice catch and thanks for the clean up!

PR #170, opened 2023-06-20
Clean up - Rename CoreSpline to PySpline

j042:
# Overview
There were duplicating typename `CoreSpline`. It means:
- in python -> `splinepy::py::PySpline`
- in cpp -> `std::shared_ptr<splinepy::splines::SplinepyBase>`

Now, in python `CoreSpline` is renamed to `PySpline` to avoid confusion.


danielwolff1:
Thanks for requesting review, I’ll take care of it at the beginning of next week since I am in a conference until wednesday  and on vacation after. :)

PR #173, opened 2023-06-23
Cu pyreader

markriegler:
# Overview
Changed two things in `cpp/splinepy/py/py_spline_reader.hpp`:
- function names to fit C++ convention (e.g. from `read_xml` to `ReadXml`) and 
- If elses in `ParseBspline` and `ParseNurbs` to switches

### DIFF
--- a/cpp/splinepy/py/init/reader.cpp
+++ b/cpp/splinepy/py/init/reader.cpp
@@ -2,6 +2,6 @@
 
 namespace splinepy::py::init {
 
-void init_reader(py::module_& m) { add_spline_reader(m); }
+void init_reader(py::module_& m) { AddSplineReader(m); }
 
 } // namespace splinepy::py::init
--- a/cpp/splinepy/py/init/reader.cpp
+++ b/cpp/splinepy/py/init/reader.cpp
@@ -2,6 +2,6 @@
 
 namespace splinepy::py::init {
 
-void init_reader(py::module_& m) { AddSplineReader(m); }
+void init_reader(py::module_& m) { add_spline_reader(m); }
 
 } // namespace splinepy::py::init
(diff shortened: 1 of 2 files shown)


cpp/splinepy/py/init/reader.cpp
```diff
@@ -2,6 +2,6 @@
 
 namespace splinepy::py::init {
 
-void init_reader(py::module_& m) { add_spline_reader(m); }
+void init_reader(py::module_& m) { AddSplineReader(m); }
```
j042, line 5: This should probably be a separaate PR. can you set it back to what it was before?

cpp/splinepy/py/py_spline_reader.hpp
```diff
@@ -15,6 +15,7 @@
 #include <Sources/InputOutput/xml.hpp>
 #include <Sources/Splines/b_spline.hpp>
 #include <Sources/Utilities/named_type.hpp>
+#include <splinepy/utils/print.hpp>
```
j042, line 18:
```suggestion

#include <splinepy/utils/print.hpp>
```

cpp/splinepy/py/py_spline_reader.hpp
```diff
@@ -40,16 +41,12 @@ class BSplineParser {
...
-  int i;
-  /// Counter variable for inner loops
-  int j;
-
   BSplineParser() {}
 
   /// @brief Convert BSpline to Python dictionary
   /// @param bspline BSpline object
-  py::dict bspline_to_dict(SplineEntry const& bspline) {
+  py::dict BsplineToDict(SplineEntry const& bspline) {
```
j042, line 48:
```suggestion
  py::dict BSplineToDict(SplineEntry const& bspline) {
```
same for the other ones. (Bspline -> BSpline)

j042 (commented):
Thanks!

cpp/splinepy/py/init/reader.cpp
markriegler: Also changed the function name in `cpp/splinepy/py/py_spline_reader.hpp` to `add_spline_reader`, as it was before.

j042 (approved):
Thank you!

PR #174, opened 2023-06-27
Add Base Spline Testcase 

deltaLuki:
# Overview
To make the test case easier to read, a general test case for splines was created inside the common.py file, which has a set of splines. Furthermore, some functions have been renamed to make the code cleaner.    
## Showcase

```
pytest tests/ --verbose
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -17,87 +17,100 @@ __all__ = [
 # b: bspline
 # n: nurbs
 
-control_points_bspline = [
-    [0, 0],
-    [0, 1],
-    [1, 1.5],
-    [3, 1.5],
-    [-1, 0],
-    [-1, 2],
-    [1, 4],
-    [3, 4],
-    [-2, 0],
-    [-2, 2],
-    [1, 5],
-    [3, 5],
-]
 
-control_points_nurbs = [
-    [-1.0, 0.0],
-    [-1.0, 1.0],
-    [0.0, 1.0],
-    [-2.0, 0.0],
-    [-2.0, 2.0],
-    [0.0, 2.0],
-]
-
-control_points_bezier = [
-    [-1.0, 0.0],
-    [-1.0, 1.0],
-    [0.0, 1.0],
-    [-2.0, 0.0],
-    [-2.0, 2.0],
-    [0.0, 2.0],
-]
+def get_2d_control_points_b_spline():
+    return [
+        [0, 0],
+        [0, 1],
+        [1, 1.5],
+        [3, 1.5],
+        [-1, 0],
+        [-1, 2],
+        [1, 4],
+        [3, 4],
+        [-2, 0],
+        [-2, 2],
+        [1, 5],
+        [3, 5],
+    ]
 
 
-control_points_splines_3d = [
-    [0.0, 0.0, 0.0],
-    [1.0, 0.0, 0.0],
-    [0.0, 1.0, 0.0],
-    [1.0, 1.0, 0.0],
-    [0.0, -1.0, 1.0],
-    [1.0, 0.0, 1.0],
-    [-1.0, 1.0, 2.0],
-    [2.0, 2.0, 2.0],
-]
+def get_2d_control_points_nurbs():
+    return [
+        [-1.0, 0.0],
+        [-1.0, 1.0],
+        [0.0, 1.0],
+        [-2.0, 0.0],
+        [-2.0, 2.0],
+        [0.0, 2.0],
+    ]
 
 
-knot_vectors_2d = [
-    [0, 0, 0, 0.5, 1, 1, 1],
-    [0, 0, 0, 1, 1, 1],
-]
+def get_2d_control_points_bezier():
+    return [
+        [-1.0, 0.0],
+        [-1.0, 1.0],
+        [0.0, 1.0],
+        [-2.0, 0.0],
+        [-2.0, 2.0],
+        [0.0, 2.0],
+    ]
 
 
-knot_vectors_3d = [
-    [0.0, 0.0, 1.0, 1.0],
-    [0.0, 0.0, 1.0, 1.0],
-    [0.0, 0.0, 1.0, 1.0],
-]
+def get_3d_control_points():
+    return [
+        [0.0, 0.0, 0.0],
+        [1.0, 0.0, 0.0],
+        [0.0, 1.0, 0.0],
+        [1.0, 1.0, 0.0],
+        [0.0, -1.0, 1.0],
+        [1.0, 0.0, 1.0],
+        [-1.0, 1.0, 2.0],
+        [2.0, 2.0, 2.0],
+    ]
+
+
+def get_knotvectors_2():
+    return [
+        [0, 0, 0, 0.5, 1, 1, 1],
+        [0, 0, 0, 1, 1, 1],
+    ]
+
+
+def get_knotvectors_3():
+    return [
+        [0.0, 0.0, 1.0, 1.0],
+        [0.0, 0.0, 1.0, 1.0],
+        [0.0, 0.0, 1.0, 1.0],
+    ]
+
 
 # query points
-q2D = [
-    [0.01, 0.01],
-    [0.01, 0.5],
-    [0.9, 0.1],
-    [0.8, 0.7],
-    [0.4, 0.99],
-]
+def get_query_points_q2D():
+    return [
+        [0.01, 0.01],
+        [0.01, 0.5],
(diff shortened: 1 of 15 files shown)


tests/common.py
```diff
@@ -74,123 +120,147 @@ def n2p2d():
...
+    )
 
 
-def all3p3d():
-    z = splinepy.Bezier(**z3p3d())
-    r = splinepy.RationalBezier(**r3p3d())
-    b = splinepy.BSpline(**b3p3d())
-    n = splinepy.NURBS(**n3p3d())
-    return [z, r, b, n]
+def bezier_3d():
```
j042, line 235: let's keep the parametric dimension info (2p3d)

tests/common.py
```diff
@@ -316,3 +386,41 @@ def are_stripped_lines_same(a, b, ignore_order=False):
...
+        splinepy.RationalBezier(),
+    ]
+
+
+def get_all_splines_as_dict_as_list():
+    return [dict_b2p2d(), dict_n2p2d(), dict_z2p2d(), dict_r2p2d()]
+
+
+class SplineBasedTestCase(unittest.TestCase):
+    def setUp(self):
```
j042, line 414: as discussed, we will use each creation as member function instead of putting it in setUp()

tests/test_bezier_extraction.py
```diff
@@ -45,7 +45,7 @@ def test_extraction(self):
     def test_extraction_matrice_bspline_3D(self):
         """Create matrices to extract splines"""
         # Init b-splines
-        bspline = c.splinepy.BSpline(**c.b3P3D)
+        bspline = self.spline_b3p3d()
```
jzwar, line 48: Why is this not called `bspline_3p3d`? This is just as confusing as it was before

tests/test_bezier_extraction.py
```diff
@@ -82,7 +82,7 @@ def test_extraction_matrices_nurbs_2D(self):
         """Create matrices to extract splines"""
 
         # Init nurbs
-        nurbs = c.splinepy.NURBS(**c.n2P2D.copy())
+        nurbs = self.spline_n2p2d()
```
jzwar, line 85: Same here

tests/test_c_contiguous.py
```diff
@@ -9,27 +9,19 @@ def test_c_contiguous_array_input(self):
...
-            c.splinepy.BSpline,
-            c.splinepy.NURBS,
-        )
-        spl_props = (c.z2P2D, c.r2P2D, c.b2P2D, c.n2P2D)
-
-        for spline_t, props in zip(spl_ts, spl_props):
-            p = props.copy()  # create shallow copy of the dict
 
+        for spl, props in zip(
+            c.get_all_spline_typs_as_list(),
```
jzwar, line 14:
```suggestion
            c.spline_types_as_list(),
```

tests/test_c_contiguous.py
```diff
@@ -9,27 +9,19 @@ def test_c_contiguous_array_input(self):
...
+            c.get_all_splines_as_dict_as_list(),
```
jzwar, line 15:
```suggestion
            c.spline_dictionaries(),
```

jzwar: Plural implies list

tests/test_c_contiguous.py
```diff
@@ -9,27 +9,19 @@ def test_c_contiguous_array_input(self):
...
+        ):
             # make f contiguous
-            f_contig_orig = c.np.asarray(p["control_points"], order="F")
+            f_contig_orig = c.np.asarray(props["control_points"], order="F")
```
jzwar, line 18: idk, in my head, props sounds like saying "proper respect", as in props, why not just use the entire word?
```suggestion
            f_contig_orig = c.np.asarray(properties["control_points"], order="F")
```

tests/test_evaluation.py
```diff
@@ -4,16 +4,12 @@
...
-        self.z2P2D = c.z2P2D.copy()
-        self.r2P2D = c.r2P2D.copy()
-        self.bspline = c.splinepy.BSpline(**self.b2P2D)
-        self.nurbs = c.splinepy.NURBS(**self.n2P2D)
-        self.bezier = c.splinepy.Bezier(**self.z2P2D)
-        self.rational = c.splinepy.RationalBezier(**self.r2P2D)
+        self.bspline = self.spline_b2p2d()
+        self.nurbs = self.spline_n2p2d()
+        self.rational = self.spline_r2p2d()
+        self.bezier = self.spline_z2p2d()
```
jzwar, line 12: This puts us in the same position as before

tests/test_evaluation.py
```diff
@@ -127,13 +123,13 @@ def test_basis_and_support(self):
         # test basis functions
         self.assertTrue(
             c.np.allclose(
-                self.bspline.basis_and_support(c.q2D)[0],
+                self.bspline.basis_and_support(c.get_query_points_q2D())[0],
```
jzwar, line 126:
```suggestion
                self.bspline.basis_and_support(c.queries_2D())[0],
```

tests/test_evaluation.py
```diff
@@ -238,7 +234,7 @@ def test_basis_function_derivatives(self):
         q2D = c.np.random.rand(10, 2)
 
         # Rational Bezier and NURBS are equivalent but use different backends
-        rational_c = self.rational.copy()
+        rational_c = self.spline_r2p2d().copy()
```
jzwar, line 237:
```suggestion
        rational_c = self.rationalbezier_2p2d().copy()
```
see above

tests/test_kv_manipulation.py
```diff
@@ -4,14 +4,10 @@
...
+class TestSplinepyKnotVectorManipulation(c.SplineBasedTestCase):
     def setUp(self):
-        self.b2P2D = c.b2P2D.copy()
-        self.n2P2D = c.n2P2D.copy()
-        self.bspline = c.splinepy.BSpline(**self.b2P2D)
-        self.nurbs = c.splinepy.NURBS(**self.n2P2D)
-        self.ref_bspline = c.splinepy.BSpline(**c.b2P2D)
-        self.ref_nurbs = c.splinepy.NURBS(**c.n2P2D)
+        self.bspline = self.spline_b2p2d()
+        self.nurbs = self.spline_n2p2d()
```
jzwar, line 10: .

tests/test_spline_type_conversion.py
```diff
@@ -18,9 +18,13 @@ def convert_and_compare_samples(spline, conversion_list):
...
 
-class SplineTypeConversionTest(c.unittest.TestCase):
+class SplineTypeConversionTest(c.SplineBasedTestCase):
     def setUp(self):
-        self.z, self.r, self.b, self.n = c.all2p2d()
+        splines = c.get_all_spline_typs_as_list()
+        self.b = splines[0]
+        self.n = splines[1]
+        self.z = splines[2]
+        self.r = splines[3]
```
jzwar, line 27: Can this be avoided? this is very `c`onfusing

jzwar (changes requested):
I left a few remarks. Also, outdated issues should be marked as **resolved**, otherwise, I get the impression that this is still being worked on! If there are still some open discussions from previous reviews, I naturally assume that there are no updates :man_shrugging:

tests/test_bezier_extraction.py
deltaLuki: I use the abbreviations given in the common.py file. But of course I can rename it.

tests/test_evaluation.py
deltaLuki: If a spline is needed multiple times in each test case, I think it makes more sense to get them from the `SplineBasedTestCase` class at the beginning of the test case and use it throughout it.

tests/test_kv_manipulation.py
deltaLuki: Here the knot vectors of the splines are edited, so the same spline must be used.

tests/test_spline_type_conversion.py
jzwar: the problem is not the assignment, though I would prefer:
```
[self.b, self.n, self.z, self.r] =  c.get_all_spline_typs_as_list()
```
however, it's the names that are unclear. I'd rather have it called. bezier, nurbs, etc. ... but still. We should not use the setup routine as functions are prone to interfere! That is the whole reason this pr exists.

PR #181, opened 2023-06-28
Implemented reading of irit files

markriegler:
# Overview
Loading irit files is now supported.

## Comment
Nltk is used to extract the control point and knot vector entries.


clemens-fricke:
Please add the `nltk` package to the dependencies. This needs to be done in the `README.md` and `setup.py` files. I am not sure if this will fix all the errors, but it should fix a few.

PR #183, opened 2023-06-29
Copy of C++ spline lead to 2 splines sharing core

jzwar:
This is more readable and avoids copy bugs. Transformation operations should always end in copies of the original, regardless of whether the type was changed or not

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/microstructure/microstructure.py
+++ b/splinepy/microstructure/microstructure.py
@@ -289,7 +289,9 @@ class Microstructure(SplinepyBase):
 
         # Prepare the deformation function
         # Transform into a non-uniform splinetype and make sure to work on copy
-        if hasattr(self._deformation_function, "bspline"):
+        if "BSpline" in self._deformation_function.whatami:
+            deformation_function_copy = self._deformation_function.copy()
+        elif hasattr(self._deformation_function, "bspline"):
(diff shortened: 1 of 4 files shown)


splinepy/microstructure/microstructure.py
```diff
@@ -289,7 +289,9 @@ def create(self, closing_face=None, knot_span_wise=None, **kwargs):
 
         # Prepare the deformation function
         # Transform into a non-uniform splinetype and make sure to work on copy
-        if hasattr(self._deformation_function, "bspline"):
+        if "BSpline" in self._deformation_function.whatami:
+            deformation_function_copy = self._deformation_function.copy()
+        elif hasattr(self._deformation_function, "bspline"):
```
clemens-fricke, line 294: What is the difference here exactly.

clemens-fricke (approved):

splinepy/microstructure/microstructure.py
jzwar: In the case a bspline is passed as a deformation function, there is no conversion. This serves only for readability.

j042:
This is an incomplete fix

jzwar:
> This is an incomplete fix

What is messing @j042

PR #184, opened 2023-06-29
Automatic volume integration

jzwar:
**Is draft because no tests yet**

# Overview
Calculate volume. Had to write this function too many times, so I decided it just be a member.

## Addressed issues
- I was tired of writing the same function over and over

## Showcase
```python
length = line_spline.integrate.volume()
area = surface_spline.integrate.volume()
volume = volume_spline.integrate.volume()
```

**Is draft because no tests yet**

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


jzwar:
If you have issues with that we can also make a `helpme`-type call

j042:
> If you have issues with that we can also make a `helpme`-type call

This is a cool function, but that sounds even cooler. Than we can later implement more integration utils. Could you start a helpme/integrate.py file?

jzwar:
> > If you have issues with that we can also make a `helpme`-type call
> 
> This is a cool function, but that sounds even cooler. Then, we can later implement more integration utils. Could you start a helpme/integrate.py file?

Done

splinepy/helpme/integrate.py
```diff
@@ -0,0 +1,125 @@
...
+    orders : array-like (optional)
+      order for gauss quadrature
+
+    Returns
+    -------
+    volume : float
+      Integral of dim-dimensional object
+    """
+    from splinepy.spline import Spline
+    from splinepy.utils.data import cartesian_product
```
j042, line 28:
```suggestion
```

splinepy/helpme/integrate.py
```diff
@@ -0,0 +1,125 @@
+import numpy as np
+
```
j042, line 2:
```suggestion

from splinepy.utils.data import cartesian_product

```
If this doesn't cause any import issues, I'd prefer this

PR #193, opened 2023-07-13
disk to disc

j042:
# Overview
spelling preference.

### DIFF
--- a/examples/create_basic_splines.py
+++ b/examples/create_basic_splines.py
@@ -21,12 +21,12 @@ if __name__ == "__main__":
     )
 
     circ = splinepy.helpme.create.circle(5)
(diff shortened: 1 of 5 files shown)


clemens-fricke:
I am not sure if this is a good change,

Disc is the British spelling and most programming is done in American English. Like color etc. 

Does it make sense to have both American and British English in our code?

I have not looked further into other word differences between a.en and b.en in our code and what version we used, this was just the first thought I had when reading this.

j042:
valid point, is there any ways to check that?

clemens-fricke:
Only manually, but since I just found out that center (a.en) is in b.en centre I am not sure if I would spot all occurrences of different spellings.

But for center we use a.en.

PR #196, opened 2023-07-19
Clean-up CPP

jzwar:
# Overview
Much of the unused, unmaintained code in bezman was stripped from the base, core functions were moved to splinepy.

## Addressed issues
* Cluttered code
* Get rid of all warnings during the build
* Faster execution by avoiding copies of points in interfaces

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar:
Couple of unresolved issues open for discussion:
- Currently online tests fail, because uffpy is locally imported (will be resolved, once either uff or funi are added as dependencies)
- `py_interfaces.hpp` is not initialized as it is directly used by `py_multi_patch`. Do we want to bind it and determine interfaces from python side? Where to place the file to keep the structure consistent? 
- MFEM is stripped of the `edges`-group. Is this temporary or can it stay like this?

Let's talk about it in person @j042

jzwar:
Maybe another option would be to integrate the `uff` routine (now that it works) into `funi`? I just don't want to let the project die.

jzwar:
This PR is on hold till all these PR are merged:
 - [ ] [uff](https://github.com/tataratat/uff/pull/6/)
 - [ ] [BSplineLib](https://github.com/tataratat/BSplineLib/pull/18)
 - [ ] [bezman](https://github.com/tataratat/bezman/pull/13)

j042 (approved):
so clean! thanks!

PR #201, opened 2023-07-24
fix #198

### LINKED ISSUE #198: Mutipatch conversion error on multiple calls
```
import splinepy as sp
import numpy as np
a = sp.Bezier(degrees=[1,1,1],control_points=sp.utils.data.cartesian_product([np.array([0,1]) for _ in range(3)]))
aa = a.copy()
aa.cps += [1,0,0]
b = sp.Multipatch(splines=[a,aa])
# First
v = b.boundary_multi_patch()
v.splines # Works fine
v = b.boundary_multi_patch()
v.splines # Error
```
The above example produces an error as the shared pointer (member of `py_multi_patch`) is not converted into the correct type. Maybe @j042 you have an idea what causes this problem?

j042:
# Overview
- multi patch keeps reference of derived object
- removes move ctor and copy members. here, all the py::* members will just be a underlying reference copy, whereas cpp type members will be copied. But they are either `shared_ptr` or vector thereof.

Note
----
This again depends on the assumption that multi patch will never have inplace changes. Some ways to enforce this is to:
1. always return shallow copy of multipatch.patches
2. add "freeze" option to splines.
3. maybe, maybe, revive splinelist and use pybind's `smart_holder`

## Addressed issues
*  #198


jzwar (approved):
Nice work,thanks! This has the additional advantage, that `ToDerived()` is also only called just once!

PR #194, opened 2023-07-17
Multipatch Extractor

jzwar:
# Overview
Create routines for multipatch class. Facilitate and accelerate multipatch plotting.

## Addressed issues
*  Plotting was very slow

## Showcase
TBD

## Checklists
* [ ] Documentations are up-to-date.
* [ ] Added example(s)
* [x] Added test(s)

PR #203, opened 2023-07-25
Split hpp cpp

j042:
# Overview
This PR splits header files and non-template implementations. As a result, `cpp` folder is split into `include` and `src` folder.

## Addressed issues
* too many files in `cpp` 

## Note
Waiting for #194. Then py_multi_patch.hpp will be properly separated.


include/splinepy/fitting/lu_solve.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <cmath>
+#include <vector>
+
+namespace splinepy::fitting {
+
+void Doolittle(const std::vector<double>& matrix,
```
jzwar, line 8: Maybe some comments? (I would also accept "next PR")

src/py/py_spline.cpp
```diff
@@ -0,0 +1,927 @@
...
+            kv_dim,
+            ").",
+            "At least",
+            n_minimal_knots,
+            "knots are expected, but",
+            nknots,
+            "were given.");
+      }
+
+      // Only use closed knot-vectors
```
jzwar, line 115:
```suggestion
```

jzwar (approved):
I guess that most of these changes are renamings and in some cases splitting of hpp and cpp. I did not read all the files in detail, so if there are any sneaky changes in the actual implementation, well played :wink: 
Thanks for putting in the effort. You are instantly rewarded with 6e4 line changes :D LGTM

include/splinepy/fitting/lu_solve.hpp
j042: next PR

PR #204, opened 2023-07-25
use ruff and update changes

j042:
# Overview
uses ruff instead of flake8. It is fast.


jzwar (approved):
LGTM, thank you. I assume that this will not change any of the former layout options so no code will be changed by applying this

PR #205, opened 2023-07-25
less explicit and split splinepy-bezier headers

j042:
# Overview
- Originally thought that it'd be a good idea to explicitly instantiate even more classes, but it seems that (at least with current setup) it doesn't make any difference. I realized that explicit base classes from bsplinelib and bezman does not contribute to less compile time, so those files are removed.  
- implementations of `splines/bezier.hpp` and `splines/rational_bezier.hpp` are separated into their own `.inl` files. Those two have circular dependency and this way, we don't really need forward declarations.

## Addressed issues
*  want less files to maintain


jzwar (approved):
I really like the new structure. It is much more intuitive. I can not say anything on the compilation times, and how those are effected. But in terms of intuitiveness, this structure beats the old one easily. LGTM, thanks for putting in the effort. I think we should prioritize the other PRs though as they will end up with a lot of merge-conflicts...

j042:
ah yes, this one goes to `split-hpp-cpp`, and that one we can wait until #194 merges. I intentionally left `py_multi_patch.hpp` untouched - `successfully rebased` message at first shot is just nice

PR #206, opened 2023-07-26
CATS-XML IO

jzwar:
# Overview
Support CATS-type xml format.

## Showcase
```
import splinepy as sp
cone = sp.helpme.create.cone(5,10,angle=180)
sp.io.cats.export("test.xml", cone)
list_of_splines = sp.io.cats.load("test.xml")
```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


splinepy/io/cats.py
```diff
@@ -0,0 +1,306 @@
...
+                )
+            n_patches = int(
+                child.attrib.get(CATS_XML_KEY_WORDS["n_patches"], 0)
+            )
+            for patch_element in child:
+                list_of_splines.append(_read_spline(patch_element))
+        else:
+            debug(f"Unused xml-keyword {child.tag}")
+    debug(f"Found a total of {len(list_of_splines)} " f"BSplines and NURBS")
+    if len(list_of_splines) != n_patches:
```
j042, line 160:
```
    if len(list_of_splines) != n_patches:
UnboundLocalError: local variable 'n_patches' referenced before assignment
```
I guess it doesn't reach l.152 ?

j042:
Have you tried a round-trip? load fails

jzwar:
> Have you tried a round-trip? load fails
There is one in the test. However, I realized, that the SplineLib example does not work because of a default root setting. I updated the code accordingly and now everything should be fine.

splinepy/io/cats.py
jzwar: Previously root was "xml" however, it is now updated to `SplineList`

j042 (approved):
thank you so much!

PR #209, opened 2023-07-27
add cmake-format, codespell, blacken-docs

j042:
# Overview
adds more pre-commit.


jzwar (approved):
Neat. I am actually impressed, that it also checks variable name spelling and not only comments (e.g. `splitted`).
LGTM, thank you

PR #213, opened 2023-08-01
fix show_options and copy() in PhysicalSpaceArray

j042:
# Overview
- `something` in `show_options` check uses `show_options.keys()`. This is required since show_options is not a dict. This is automatically taken away by ruff. To support former syntax, `__contains__` will be implemented in gustaf
- `PhysicalSpaceArray.copy()` is supposed to return a `np.array` copy, but it wasn't, which caused no-update issues in `show_spline_data.py` example. fixed now.

## Addressed issues
*  error in  `examples/show_spline_data.py`

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


splinepy/utils/data.py
```diff
@@ -648,12 +648,13 @@ def __setitem__(self, key, value):
         """
         if isinstance(value, SplineDataAdaptor):
             self._saved[key] = value
-        elif "PyMultiPatch" in str(type(value).__mro__):
+        elif "PyMultipatch" in str(type(value).__mro__):
```
jzwar, line 651: nice catch

jzwar (approved):
LGTM - tested it locally works fine. Thank you for the fast bug fix

PR #217, opened 2023-08-07
Change numbering system of interfaces

jzwar:
# Overview
Up to this point, interfaces were stored in the `mrng`-style. This had two major downfalls. 
1. The adjacent face ID had to be calculated separately
2. The adjacent face ID was not uniquely identifiable if more than one intersection existed with the same patch.

These issues can be solved if the interfaces do not refer to the neighboring patch ID, but rather to the neighboring face ID in the global counting system.

## Addressed issues
*  Issue was communicated via Rocket.Chat

## Checklists
* [ ] Added example(s)
* [x] Added test(s)

### DIFF
--- a/splinepy/io/gismo.py
+++ b/splinepy/io/gismo.py
@@ -237,42 +237,28 @@ def export(
 
     interface_data = ET.SubElement(multipatch_element, "interfaces")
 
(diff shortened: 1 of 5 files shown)


j042 (approved):
Thank you so much! Great change

PR #221, opened 2023-08-09
Add function CheckConformity for multipatch

deltaLuki:
# Overview

## Addressed issues
*  Issues addressed

## Showcase
```
import splinepy

new, feature = splinepy.tataratat()
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar:
No commits in 6 weeks, is this still being worked on?

deltaLuki:
> No commits in 6 weeks, is this still being worked on?

Yes, it's still a work in progress. Testing with more than two patches still fails. I also had to change some functions.

j042:
can you push your progress?

jzwar:
As discussed offline, your commit history should not only show results but also document your steps to get there. This way, we also see that this is still in active development. I did not expect this task to take so long, to be honest. Please do not hesitate to discuss problems, maybe we can find a solution together.

src/py/py_multipatch.cpp
```diff
@@ -1627,8 +1793,16 @@ void init_multipatch(py::module_& m) {
            py::arg("checK_control_mesh_resolutions"),
            py::arg("nthreads"))
       .def("fields", &PyMultipatch::GetFields)
-      //.def("", &PyMultipatch::)
-      ;
+      .def("orientations",
```
jzwar, line 1796:
```suggestion
      .def("interface_orientations",
```

src/py/py_multipatch.cpp
```diff
@@ -671,68 +671,242 @@ void GetBoundaryOrientation(
...
+    const int& start_patch_id = orientations_ptr[i * n_entries_per_line + 0];
+    const int& start_face_id = orientations_ptr[i * n_entries_per_line + 1];
+    const int& end_patch_id = orientations_ptr[i * n_entries_per_line + 2];
+    const int& end_face_id = orientations_ptr[i * n_entries_per_line + 3];
+    const int* alignment_ptr = &orientations_ptr[i * n_entries_per_line + 4];
+    const int* orientation_ptr =
+        &orientations_ptr[i * n_entries_per_line + 4 + param_dim];
+
+    // get coordinates pointer from control point
+    const auto start_control_points =
```
jzwar, line 833: Here you have to check if the spline is rational or not! Otherwise the function will not work...

src/py/py_multipatch.cpp
```diff
@@ -671,68 +671,242 @@ void GetBoundaryOrientation(
...
+                                    const std::vector<int> cmr,
+                                    const int coordinate_id_start_patch,
+                                    const int direction) -> int {
+    if (face_id > 1) {
+      return cmr[direction % 2] - 1;
+    }
+    return coordinate_id_start_patch;
+  };
+
+  for (int i{}; i < (core_patches_.size() - 1); i++) {
```
jzwar, line 823: This should not be about the patches but about patch connections

src/py/py_multipatch.cpp
```diff
@@ -671,68 +671,250 @@ void GetBoundaryOrientation(
...
+                                i_face_ctp);
+        const int ctps_id_start_patch =
+            coord_id_to_glob_id(cmr_start,
+                                coordinate_id_start_patch,
+                                start_face_id);
+        // transformation start -> end
+        std::vector<int> coordinate_id_end_patch(param_dim);
+        for (int p_dim{}; p_dim < param_dim; p_dim++) {
+          const int& p_axis_end = alignment_ptr[p_dim];
+          if (end_face_id == p_axis_end) {
```
jzwar, line 883:
```suggestion
          if (end_face_id / 2 == p_axis_end) {
```
Here is an error

src/py/py_multipatch.cpp
```diff
@@ -671,68 +671,250 @@ void GetBoundaryOrientation(
...
+          return false;
+        }
+      }
+    }
+    return true;
+  };
+
+  // @Lukas NThreadExecution please
+  check_conformity_of_interface(0, orientations_.shape(0));
+  return true;
```
jzwar, line 917:
```suggestion
  return check_conformity_of_interface(0, orientations_.shape(0));
```

jzwar (changes requested):
Please try these changes. However  currently, the function is still inconsistent as it throws an exception if the first step fails and returns false if the second test fails

tests/test_multipatch.py
```diff
@@ -76,6 +76,100 @@ def test_interfaces(self):
             ).all()
         )
 
+    def test_checkConformityTwoPatches_2d(self):
```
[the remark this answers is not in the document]
jzwar: Resolved

jzwar:
There are still some open issues:
- [ ] Rational splines need "weighted control points"
- [ ] Tests are not complete, we need all possible orientations with and without conformity to verify that the function works
- [ ] Different "Rationality" of two adjacent splines can occur, then weights must be equal for the entire spline
- [ ] check_conformity needs proper python binding with documentation and default values

Optional extensions:
- [ ] Extend plotting options to show interfaces as spheres/ points wherever an interface is detected to make it easier for the user to see connections (maybe this is something @j042 can help provide?)

deltaLuki:
> There are still some open issues:
> 
> * [ ]  Rational splines need "weighted control points"
> * [ ]  Tests are not complete, we need all possible orientations with and without conformity to verify that the function works
> * [ ]  Different "Rationality" of two adjacent splines can occur, then weights must be equal for the entire spline
> * [ ]  check_conformity needs proper python binding with documentation and default values
> 
> Optional extensions:
> 
> * [ ]  Extend plotting options to show interfaces as spheres/ points wherever an interface is detected to make it easier for the user to see connections (maybe this is something @j042 can help provide?)

Check rational splines: Should the values of the weights also be checked? (e.g. whether a weight is negative)

jzwar:
> Check rational splines: Should the values of the weights also be checked? (e.g. whether a weight is negative)

No this will not be necessary, as spline constructors already provide these checks. Here we only need to check:
1. if both the start as well as the end spline are rational:
   - weights are equal
   - weighted control points are equal (which by consequence means that ctps are also equal)
2. if both splines are non-rational:
   - ctps are equal
3. if one spline is rational and the other spline is non-rational:
   - control-points (that is: ctps/weight) are equal
   - all weights are the same value

j042:
any updates?

deltaLuki:
> any updates?

I think that there is an bug in creating the orientations when the axes of two patches are rotating by 180 degrees. 
#269

j042:
status?

deltaLuki:
> status?

From my point of view (see comment - checkbox of @jzwar ), it is finished and can be checked.

j042:
can you rebase and fix errors?

deltaLuki:
I have an issue with the pipeline 'docs/minimal_explicit_build_and_docs.' Initially, 'clang++-12' couldn't be found by the GitHub runner. Because of this, I increased the version to 14, but now I'm encountering a different error with something like 'navigation_with_keys.' Is there a common problem with the CI here, or is there an already working pipeline? @j042

j042:
> I have an issue with the pipeline 'docs/minimal_explicit_build_and_docs.' Initially, 'clang++-12' couldn't be found by the GitHub runner. Because of this, I increased the version to 14, but now I'm encountering a different error with something like 'navigation_with_keys.' Is there a common problem with the CI here, or is there an already working pipeline? @j042

rebase

j042:
please rebase instead

deltaLuki:
> please rebase instead

After rebasing the pipeline still fails

j042:
I see. maybe you can create a tiny separate PR to fix that?

j042:
closing this due to inactivity. Feel free to reopen

PR #223, opened 2023-08-10
Tiny bugfix

jzwar:
# Overview
There was an issue in `Multipatch->set_boundary()`, where an exception was shadowed by another catch block which led to unhelpful output messages.


jzwar:
Thanks to @kkCATS for pointing that out

PR #227, opened 2023-08-25
Binary Gismo IO

jzwar:
# Overview
This feature provides binary io for gismo files, which is important, as tiny changes can already affect the overall structural sensitivity.

## Addressed issues
*  Sensitivity issues on optimization routine

## Showcase
```
import splinepy
box = splinepy.helpme.create.box(10,10,20).nurbs
box_mp = splinepy.Multipatch([box])
splinepy.io.gismo.export(box_mp, as_binary=True)
```

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


splinepy/utils/data.py
```diff
@@ -283,7 +284,7 @@ def enforce_contiguous(array, dtype=None, asarray=False):
             return array
         return np.ascontiguousarray(array, dtype=dtype)
 
-    if asarray and isinstance(array, (list, tuple)):
+    if asarray and isinstance(array, (list, tuple, splinepy_core.KnotVector)):
```
j042, line 287: nice! where was this causing problem?

.pre-commit-config.yaml
```diff
@@ -65,7 +65,7 @@ repos:
   rev: v2.2.5
   hooks:
   - id: codespell
-    args: [-w]
```
j042, line 68: is this relevant? if not rm?

j042 (commented):
thanks! any idea when this will be available from gismo?

jzwar:
> thanks! any idea when this will be available from gismo?

I hope very soon!

.pre-commit-config.yaml
jzwar: Yes I think this can be removed now that the files are stored externally

splinepy/io/gismo.py
```diff
@@ -190,9 +220,8 @@ def export(
     export_fields : bool
       Export fields to separate files ending with field<id>.xml, e.g.,
       filename.xml.field1.xml
-    collapse_fields : cool
-      Only valid if export_fields is True, writes all fields in the same file
-      by collapsing the control points, requires conformity (not checked)
+    as_base64 : bool (False)
+      If set, coordinates, weights and knot-vectors are exported in B64 ecoding
```
j042, line 224:
```suggestion
      If set, coordinates, weights and knot-vectors are exported in B64 encoding.
      Available in gismo, once gismo/gismo#634 merges.
```
I think with this note, we can merge this. You could update readme too if you want

j042 (approved):
good stuff, thanks!

PR #228, opened 2023-08-25
add mfem dof_mapping

j042:
# Overview
this PR removes `mfem_index_mapping` and adds `dof_mapping` instead. This supports both 2d and 3d using `multi_index`

## Addressed issues
*  3d missing

## Next PR
* tests - io round trip
* extend io for 3d single patch
* multi patch load

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/io/mfem.py
+++ b/splinepy/io/mfem.py
@@ -7,7 +7,12 @@ import gustaf as gus
 import numpy as np
 
 # single function imports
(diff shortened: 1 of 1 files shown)


jzwar (approved):
Amazing, so intuitive, not at all `c`onfusing... /s

LGTM, much better than the version that we had before... Thanks

PR #229, opened 2023-08-25
add flatten

j042:
# Overview
this function was in `mfem_index_mapping`. This PR adds modified version of it to utils.


jzwar (approved):
LGTM, nice work. However, I don't see what use case you need this for

j042:
yeah, was a function inside a function that seemed worth saving. Maybe it is better to move this to somewhere like `gus.utils.arr` or `gus.helpers.data`

j042:
I will move this to gustaf

PR #231, opened 2023-08-30
cmake cleanup

j042:
# Overview
- separate python binding and python module target
- export python binding target (splinepy_python)
- ensure BSplineLib target
- find_dependency in config
- update BSplineLib commit

maybe we can include those static libraries in distribution

## Addressed issues
*  linking the library to other project wasn't working as expected

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar (commented):
Is this related to the mail commentary by Matthias?

j042:
> Is this related to the mail commentary by Matthias?

no, I was trying to link splinepy to use `PySpline` in other library and realized it wasn't working as expected

jzwar (approved):
LGTM, thanks for dealing with all the building stuff

Issue #233, opened 2023-08-31
`mfem.export_cartesian()` causes `std::bad_malloc` 


j042:
multipatch export for mfem causes `std::bad_malloc` if you try to read it with mfem. any idea @jzwar ?

jzwar:
That could be a conversion error. Or most likely I messen up when I removed edges. Did you take a look at the export file? Is there maybe some redundant edge info that I forgot to delete?

j042:
i think there's a keyword missing,
```
edges
0
```

j042:
fixed with #238

PR #234, opened 2023-08-31
turn knot vectors to true list before export

j042:
# Overview
we don't want to export __repr__ of KnotVector


jzwar:
This is related to #233 I guess

j042:
not related, this is just for a single patch

jzwar (approved):

PR #237, opened 2023-08-31
Add gmsh io

Roxana-P:
# Overview
Splines are a great parametrization for easy and complex shapes.  Isogeometric analysis allows to calculate direct on spline parametrization.  However, not every simulation is equipped with the options to use splines for analysis. Thus, it can be necessary to create computational  meshes from splines. In this example, we demonstrate how to use gmsh  for meshing splinepy splines. In this sense, a connection to gmsh can allow connection between the users of these OpenSource tools. 

## Addressed issues
*  None

## Showcase
```
import splinepy
import gmsh

gmsh.initialize()
dtag, point_tags = splinepy.io.gmsh.spline_to_gmsh(splinepy.helpme.create.disk(1, angle=180))
    
```

## Checklists
* [x] Added example(s)
* [x] Added test(s)

screenshot

### DIFF
--- a/examples/gmsh_from_spline.py
+++ b/examples/gmsh_from_spline.py
@@ -1,0 +1,90 @@
+import gmsh
+
+import splinepy
+
+try:
+    import gustaf as gus
+
+    gus_exists = True
+except ModuleNotFoundError:
+    gus_exists = False
+
+
+if __name__ == "__main__":
+    """Splines are a great parametrization for easy and complex shapes.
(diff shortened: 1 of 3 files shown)


jzwar:
Can you please add testing and make `gmsh` import optional so that we don't have another dependency?
Also the `io`-module usually only provides `export` and `load` functions. It seems like you only show how to export in the example. Is there a reason for that? Can you make the spline export more general?

j042:
I'll close this for now. Feel free to reopen once `export` and `load` is ready

PR #240, opened 2023-09-01
fix get cps for multipatch

j042:
# Overview
fix wrong nthread iteration


jzwar:
That is a crucial bugfix. As for avoiding those in the future: is there a way to execute the test suite both in serial **and** in parallel?

jzwar (approved):
Thanks for fixing this! lgtm

PR #242, opened 2023-09-06
Fix/update plant kdtree

j042:
# Overview
`PlantNewKdTree()` would return early if a tree exists with same resolution. This PR removes this behavior, since control points support real in-place updates and after the updates, trees won't be up-to-date.

Also, updates napf and adds support for concurrent k-d tree build

## Addressed issues
*  outdated sample trees

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar (approved):
Patiently waiting for Greville abscissae, but till then I am fine with this. Maybe next PR?

j042:
yes!

PR #243, opened 2023-09-07
add same example using sparse matrix

j042:
# Overview
Same example, but with sparse matrix. Seems to be 2-3 times faster. I've tried some other variants to set zeros, but nothing seems to be beat simple 0 multiplication.


jzwar (approved):
Nice work, love it! LGTM

PR #246, opened 2023-09-08
Bugfix - Multipatch

jzwar:
# Overview
2 issues:
1. wrong order of arguments in pyMultipatch constructor
2. add fields in gismo output was still expecting lists of lists


j042 (approved):
good catches!

PR #252, opened 2023-09-18
fix unused function warning

j042:
# Overview
- fixes unused function warning use by making the function `static inline`. 
- adds `Werror` to `--enable_warning`


jzwar (approved):
Nice, this was already bothering me a lot. As discussed, please add `-Werror`

PR #268, opened 2023-10-03
create new cp pointer if size changes

j042:
# Overview
Previously, cp pointers are only created once and kept throughout the life of a spline. When spline has modifications through knot insertions or degree elevations, this means that the pointers aren't updated causing syncing error. This PR fixes this behavior by comparing size/length of control points before returning already existing pointers.

## Addressed issues
*  control point pointers weren't up-to-date

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar (approved):
LGTM, thank you for the update! Test looks fine. How would you feel about adding a quick check function in splinepy base like `control_point_pointers_synced`.

j042:
> LGTM, thank you for the update! Test looks fine. How would you feel about adding a quick check function in splinepy base like `control_point_pointers_synced`.

3 comments:
1. Thanks for your precious time right now ;)
2. Maybe a check function for PhysicalSpaceArray?
3. Better yet, let's go contiguous at backend?

PR #275, opened 2023-10-12
Add: Move configurations from precommit to  pyproject.toml.

clemens-fricke:
# Overview
Move the configuration of the tools used in our pre-commit to the `pyproject` file. This will enable us to have more consistent results running these tools outside the pre-commit environment.

The tools will look for a possible configuration in the `pyproject` file therefore, if we configure them there the configuration is consistent between different invocations of these tools.

## Addressed issues
*  configuration differs when running a tool defined in the pre-commit on its own without invoking pre-commit.

## Still existing issue
* Configuration can only be read in if the package `tomli` is installed. I have added the need for this package to the pre-commit configuration, but if a tool is invoked on its own, the user might (depended on the python version/package is installed due to other dependencies) need to install the package own their own.

## Checklists
 * The only thing is that dependent on the python version the configuration only gets picked up if `tomli` is installed so that the `pyproject.toml` can be read in.
* [x] Added example(s)
 * No need
* [x] Added test(s)
 * No need

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
   rev: "23.9.1"
   hooks:
   - id: black
-    args: [--line-length=79]
+    additional_dependencies: [tomli]
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.0.292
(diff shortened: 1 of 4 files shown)


j042 (approved):
thank you! can I gently ask you for a follow up PR similar to tataratat/gustaf#162 (ruff ext)

PR #276, opened 2023-10-12
tparam-free proximity

j042:
# Overview
This PR contributes to the long term goal of removing para_dim/dim from template parameters.
Proximity was taking SplineType as a template parameter.
Then it uses kParaDim and kDim to create std::arrays, nested std::arrays, and kdtree.
A solution for std::arrays is to add an Array class - this results updates in utils/arrays.hpp
As for k-d tree, turns out,  dim is not a strict template parameter. So I've updated napf to accordingly.
This should also reduce some compile tile.

## TODO
- split .cpp file

## Note
- currently, `clipped` info is not used. It used to be part of solver's skip_mask.

## Addressed issues
*  Missing flexibility in Proximity

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/include/splinepy/utils/grid_points.hpp
+++ b/include/splinepy/utils/grid_points.hpp
@@ -176,6 +176,8 @@ private:
 /// Values are kept as int since they come as int from python
 class CStyleArrayPointerGridPoints {
 public:
+  std::vector<double> step_size_;
+
   CStyleArrayPointerGridPoints() = default;
   CStyleArrayPointerGridPoints(const int dim,
                                const double* bounds,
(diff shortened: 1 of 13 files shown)


include/splinepy/utils/grid_points.hpp
```diff
@@ -195,6 +197,8 @@ class CStyleArrayPointerGridPoints {
     entries_.assign(dim, std::vector<double>{});
     resolutions_.clear();
     resolutions_.reserve(dim);
+    step_size_.clear();
+    step_size_.reserve(dim);
```
j042, line 201: Resize

danielwolff1:
Thanks a lot for changing this. I took a quick glance at most changes, checking the array implementation most thoroughly and to me it looks good!

The only remark I would have (but I guess you did this on purpose): I'm personally not a big fan of statements like `assert(pointer)`. Instead, I would always go for the -- in my opinion -- more explicit version `assert(pointer != nullptr)`.
But that's not a hard requirement but just a matter or personal style I guess.

Good job, thanks an again for the changes! 🚀

jzwar (approved):
LGTM, did not have the time to review all files, but I put a lot of faith and trust into your work, so feel free to merge <3

PR #278, opened 2023-10-13
make CStyleArrayGridPoints GridPoints

j042:
# Overview
This PR removes `GridPoints` - It was only used to extract hyperplane ids.  CStyleArrayGridPoints is renamed to GridPoints. 

## Addressed issues
*  dim as template parameter

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/include/splinepy/proximity/proximity.hpp
+++ b/include/splinepy/proximity/proximity.hpp
@@ -43,7 +43,7 @@ protected:
   const splinepy::splines::SplinepyBase& spline_;
 
   // kdtree related variables
-  splinepy::utils::CStyleArrayPointerGridPoints grid_points_;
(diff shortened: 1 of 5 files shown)


jzwar (approved):
LGTM.
I am not very happy with the name `IDsOnHyperPlane`, as it is not clear, what the function is intended for. Also: We have so many local definitions of local to global referring to multi-ID to global-ID. Is this something for utils?

j042:
we can add something like [this](https://numpy.org/doc/stable/reference/generated/numpy.ravel_multi_index.html) and [this](https://numpy.org/doc/stable/reference/generated/numpy.unravel_index.html).

PR #285, opened 2023-10-31
Updated greville argument list

jzwar:
# Greville Abscissae ++

When computing the determinant projection of a (non-rational) spline, one obvious choice for evaluation and comparison is greville abscissae, as they ensure, that all supports are sufficiently sampled. This is done, to ensure a solution for the linear system of equations. However, for $C^{-1}$ splines, there are two greville points on the high-multiplicity knots, one meant for each side of the knot line. As it is unsure on what side to evaluate this query, both points will evaluate on the  $\xi^{-}$ side.

This PR will filter duplicates out, to ensure that the supporting element is unique, by taking the mean with neighboring elements.

In order to do so, a new key has been added and also greville_abscissae lost its property status
```
spline.greville_abscissae(allow_duplicates=False)
```

## Checklists
* [X] Documentations are up-to-date.
* [ ] Added example(s)
* [X] Added test(s)


jzwar:
This is related to the work of @FSchwar

danielwolff1 (approved):
Thanks for the changes, this is very understandable and well solved in my opinion!

Thanks for updating the corresponding test as well, from my side this can be merged! 🚀

src/py/py_spline.cpp
```diff
@@ -304,42 +304,27 @@ py::array_t<double> PySpline::ParametricBounds() const {
...
-    }
-    offset *= cmr[i];
+    // Calculate Abscissae
+    py::array_t<double> greville_abscissae(cmr[i]);
+    double* greville_abscissae_ptr =
+        static_cast<double*>(greville_abscissae.request().ptr);
+    Core()->SplinepyGrevilleAbscissae(greville_abscissae_ptr,
+                                      i,
+                                      allow_duplicates);
+    greville_abscissae.resize({cmr[i]});
```
j042, line 323: Same size resize?

include/splinepy/splines/helpers/properties.hpp
```diff
@@ -103,6 +104,26 @@ inline void GetGrevilleAbscissae(const SplineType& spline,
...
   }
+
+  // There can be no duplicates for bezier types, but if C^(-1), duplicates in
+  // the knot vector result in duplicate greville abscissae which can lead to
+  // problems in further computations, solved by mean filtering with neighbors
+  if constexpr (SplineType::kHasKnotVectors) {
+    if (!allow_duplicates) {
+      double previous_knot{greville_abscissae[1]};
+      for (int j{2}; j < cmr - 1; ++j) {
+        if (std::abs(previous_knot - greville_abscissae[j]) < 1e-9) {
```
j042, line 115: this is a bit too big and fixed tolerance is not great.

j042 (changes requested):
something like duplicate_tolerance and only use/recompute it if they are positive?

src/py/py_spline.cpp
jzwar: I thought this was required to fix shape in buffer. I'll erase that

include/splinepy/splines/bezier.hpp
```diff
@@ -124,8 +124,15 @@ class Bezier : public splinepy::splines::SplinepyBase,
...
-                                         const int& i_para_dim) const;
+  /// @param[in] allow_duplicate  if true, two greville abscissae can be equal,
+  ///                             made to comply with C^(-1) splines
+  /// @param[in] duplicate_tolerance difference between two greville abscissae
+  ///                                to be considered equal
+  virtual void
+  SplinepyGrevilleAbscissae(double* greville_abscissae,
+                            const int& i_para_dim,
+                            const bool& allow_duplicates,
+                            const double& duplicate_tolerance) const;
```
j042, line 135: We don't need `allow_duplicate` as a parameter

include/splinepy/splines/helpers/properties.hpp
```diff
@@ -103,6 +111,27 @@ inline void GetGrevilleAbscissae(const SplineType& spline,
       greville_abscissae[j] = static_cast<double>(inv_factor * j);
     }
   }
+
+  // There can be no duplicates for bezier types, but if C^(-1), duplicates in
+  // the knot vector result in duplicate greville abscissae which can lead to
+  // problems in further computations, solved by mean filtering with neighbors
+  if constexpr (SplineType::kHasKnotVectors) {
+    if (duplicate_tolerance > 0) {
```
j042, line 119:
```suggestion
    if (duplicate_tolerance > 0.0) {
```

j042 (approved):
Thank you, this is great

PR #288, opened 2023-11-06
Update BSplineLib - remove dimensionality from template parameter

j042:
# Overview
Removes dimensionality from BSpline family. This PR mainly contains updates to accommodate BSplineLib's updates to use contiguous arrays as `VectorSpace`. This means we can just take buffer info from numpy arrays and use it to create a bspline. That requires us to keep the original numpy array alive for BSplines. Other than taking out `SplineType::kDim`, everything else says the same.

Avoiding copies for evaluation query/return values and switching from std::vectors to custom buffer array gave some performance boost. 

## Addressed issues
* difficult to compile
* too many copies

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar (approved):
Thank you for the quick in-person overview of the changes. Everything looks good and as the tests do not fail, I am fine with all the changes, even if I haven't seen all the lines in detail

PR #290, opened 2023-11-07
use `scikit-build-core`

### LINKED ISSUE #182: Deprecation warning in build process
Running the build on splinepy prints the following warning (Ubuntu, python 3.9)
```
<path>/site-packages/setuptools/command/easy_install.py:144: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
<path>/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
```
Is this an issue with my installation or a common problem?

j042:
# Overview
Updates/upgrades builds. fixes #182. More options available [here](https://scikit-build-core.readthedocs.io/en/stable/index.html)

## Addressed issues
*  setup.py is replaced by pyproject.toml

## Showcase
```
pip install -e. --config-settings=cmake.args=-DSPLINEPY_MORE=OFF
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar:
Did we just end support for python 3.6? Also 3.12 came out in october. It is missing in some of the files if I understand correctly

pyproject.toml
```diff
@@ -1,10 +1,54 @@
...
+urls = {Homepage = "https://github.com/tataratat/splinepy"}
+classifiers=[
+    "Development Status :: 2 - Pre-Alpha",
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
jzwar, line 22: Why no 3.12?

.github/workflows/test_full.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       SPLINEPY_GITHUB_ACTIONS_BUILD: True
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", "3.11"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
```
jzwar, line 17: 3.12 please

jzwar (approved):
LGTM :muscle:

pyproject.toml
j042: good .

j042:
I'll leave out 3.12 for now and take care of some other issues and extend it in next PR. It seems that some dependencies aren't ready for this yet and I'd need some time to update those.

pyproject.toml
j042: next PR

.github/workflows/test.yml
```diff
@@ -27,20 +27,9 @@ jobs:
...
-    - name: build splinepy macos
-      if: matrix.os == 'macos-latest'
-      run: |
-        python3 setup.py develop --minimal --debug --enable_warning
-
-    - name: build splinepy windows
-      if: matrix.os == 'windows-latest'
-      run: |
-        python3 setup.py develop --minimal --enable_warning
+        pip install . -v --config-settings=cmake.args="-DSPLINEPY_MORE=OFF;-DSPLINEPY_ENABLE_WARNINGS=ON" --config-settings=cmake.build-type="Debug"
```
clemens-fricke, line 32: So the `--minimal` option is now `-BSPLINEPY_MORE=OFF`.

Should we maybe start to invert the functionality that the option is by default `OFF` and can be set `ON` if the larger build is necessary? 
How useful is it for a normal developer/user to have the full build and not only the minimal?

Or has the problem with the incredibly high RAM and CPU usage for the full build been mitigated?

.github/workflows/test_full.yml
clemens-fricke, line 17: Above in the build workflow we already test and build the package already with `py3.12`.

CONTRIBUTING.md
```diff
@@ -10,7 +10,7 @@ git clone git@github.com:<path-to-your-fork>
 cd splinepy  # or <forkname>
 git submodule update --init --recursive
 git checkout -b new-feature0
-python3 setup.py develop --minimal --debug
+pip install -e. --config-settings=cmake.args=-DSPLINEPY_MORE=OFF --config-settings=cmake.build-type="Debug"
 ```
 `--minimal` flag builds splines upto 3D (both parametric and physical dimensions), and that way we can reduce compile time. `--debug` build also reduces compile time. We are experimenting with the ways to reduce compile time during development. Let us know if you have a great idea!
```
clemens-fricke, line 15: `--minimal` is replaced by `--config-settings=cmake.args=-DSPLINEPY_MORE`

And as mentioned above already should we think about making the `-DSPLINEPY_MORE` argument by default to `OFF` again?


```suggestion
`--config-settings=cmake.args=-DSPLINEPY_MORE` build argument builds splines up to 3D (both parametric and physical dimensions), and that way we can reduce compile time. `--config-settings=cmake.build-type="Debug"` build also reduces compile time. We are experimenting with ways to reduce compile time during development. Let us know if you have a great idea!
```

README.md
```diff
@@ -22,7 +22,7 @@ In addition to the aforementioned compilers, this requires a cmake3.16+. If you
 git clone git@github.com:tataratat/splinepy.git
 cd splinepy
 git submodule update --init --recursive
-python3 setup.py develop
+pip install -e.
```
clemens-fricke, line 25:
```suggestion
pip install -e .
```

tests/common.py
```diff
@@ -309,16 +309,22 @@ def to_tmpf(tmpd):
     return os.path.join(tmpd, "nqv248p90")
 
 
-def are_splines_equal(a, b):
+def are_splines_equal(a, b, print_=True):
```
clemens-fricke, line 312: Should we not try to reduce the printed output? This function is only for testing but still, I would normally expect options as these to be `false` by default.
```suggestion
def are_splines_equal(a, b, print_=False):
```

clemens-fricke (approved):

.github/workflows/test.yml
j042: > So the `--minimal` option is now `-BSPLINEPY_MORE=OFF`.
> 
> Should we maybe start to invert the functionality that the option is by default `OFF` and can be set `ON` if the larger build is necessary? How useful is it for a normal developer/user to have the full build and not only the minimal?

I thought about that, but I wanted to keep the  same as distribution build. I am playing with the idea to add a `scripts` or `tools` directory to gather some commands. This modern approach is log

> 
> Or has the problem with the incredibly high RAM and CPU usage for the full build been mitigated?

No. Definitely less, as explicit build of bsplines has significantly reduced. But I think nothing changed for Beziers.

.github/workflows/test_full.yml
j042: I thought I removed it?

j042: Nope, good catch!

Issue #270, opened 2023-10-03
Importing numpy leads to ModuleNotFoundError


SeRealMF:
When executing _npz.py_ in _splinepy/io_ a ModuleNotFoundError will be raised. 

This is because of the following line of code:
`import numpy as np`

Numpy is using json but because of the fact that json.py is another file in the _splinepy/io_ folder, it will therefore execute this file instead.
This leads to the following error message:
```
Traceback (most recent call last):
  File "...\splinepy\splinepy\io\npz.py", line 11, in <module>
    import numpy as np
  File "...\AppData\Local\python\mu\mu_venv-38-20221219-011508\lib\site-packages\numpy\__init__.py", line 140, in <module>
    from . import core
  File "...\AppData\Local\python\mu\mu_venv-38-20221219-011508\lib\site-packages\numpy\core\__init__.py", line 9, in <module>
    from numpy.version import version as __version__
  File "...\AppData\Local\python\mu\mu_venv-38-20221219-011508\lib\site-packages\numpy\version.py", line 3, in <module>
    from ._version import get_versions
  File "...\AppData\Local\python\mu\mu_venv-38-20221219-011508\lib\site-packages\numpy\_version.py", line 7, in <module>
    import json
  File "...\splinepy\splinepy\io\json.py", line 10, in <module>
    from splinepy import settings
ModuleNotFoundError: No module named 'splinepy'
```

j042:
This is very rare use-case. In general, I'd like to add leading underscore to module local imports, then these things shouldn't happen.

clemens-fricke:
I was not able to reproduce this error at all. Even with a completely clean environment.

SeRealMF:
I was successfully able to reproduce the error, though only on Windows, not on Linux. There may also be many other factors that influence whether the error appears or not.

j042:
> This is very rare use-case. In general, I'd like to add leading underscore to module local imports, then these things shouldn't happen.

Would still happen with json as it has the same name as built-in module. For this one, we will have to add a trailing underscore. You can add this to #296.

clemens-fricke:
After merging #309, does this issue persist? Could you check this @SeRealMF and close this issue if the problem is fixed?

SeRealMF:
The issue still persists, even with the latest changes. @j042 mentioned a trailing underscore, though the issue seems resolved in my opinion with a leading underscore of _json.py_. No errors occur when the file is named __json.py_.  
[...]

PR #296, opened 2023-11-11
Leading underscore to module imports

SeRealMF:
# Overview
Leading underscores were added to the module imports to make them private. This means that they are not accessible from outside the module and can only be used within the module itself.

## Addressed issues
[#270](https://github.com/tataratat/splinepy/issues/270)
## Showcase
```
import splinepy as _splinepy

new, feature = _splinepy.tataratat()
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,13 @@
-import os
-import platform
-import re
-import subprocess
-import sys
-from pathlib import Path
+import os as _os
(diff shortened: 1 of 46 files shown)


SeRealMF:
Should have not been linked to previous commits.

SeRealMF:
See #309

PR #309, opened 2023-11-16
Feature: leading underscore to module imports

SeRealMF:
# Overview
Leading underscores were added to the module imports to make them private. This means that they are not accessible from outside the module and can only be used within the module itself.

## Addressed issues
[#270](https://github.com/tataratat/splinepy/issues/270)
## Showcase
```
import splinepy as _splinepy

new, feature = _splinepy.tataratat()
```

### DIFF
--- a/tests/test_creator.py
+++ b/tests/test_creator.py
@@ -1,4 +1,5 @@
 import numpy as np
+
 import splinepy
 
 try:
--- a/tests/test_creator.py
+++ b/tests/test_creator.py
@@ -1,4 +1,5 @@
 import numpy as np
+import splinepy
 
 try:
     from . import common as c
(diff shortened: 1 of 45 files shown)


tests/test_creator.py
```diff
@@ -1,5 +1,7 @@
 import numpy as np
 
+import splinepy
+
```
j042, line 4:
```suggestion
```

tests/test_creator.py
```diff
@@ -94,7 +96,7 @@ def test_create_revolution(self):
 
         # Expect Failure - not a spline
         with self.assertRaises(NotImplementedError):
-            c.splinepy.spline.helpme.create.revolved([4])
+            splinepy.helpme.create.revolved([4])
```
j042, line 99:
```suggestion
            c.splinepy.helpme.create.revolved([4])
```

j042 (commented):
Thanks a lot!
Imports within the functions don't really need leading underscore - they are mostly in `splinepy/helpme/create.py`

j042 (approved):
I think we can leave function-local-import also this way for now. They are mainly for type checking and we can try to avoid in other PR. Thanks again!

This will cause some conflicts on other PRs (@jzwar, @FSchwar, @clemens-fricke), but other way around would cause bigger conflicts. So, I will merge this first

j042:
> When will this be merged? Is anything missing still?

right now

PR #310, opened 2023-11-17
Fix #239

### LINKED ISSUE #239: inplace operations on all cps causes update
I think it has something to do with python `__setitem__` syntax, but it'd be nice if we can avoid it

j042:
# Overview
This PR tries to remove some unnecessary copies

## Addressed issues
*  fixes #239

## Checklists
* [ ] Added example(s)
* [x] Added test(s)

### DIFF
--- a/.github/workflows/test_wheels.yml
+++ b/.github/workflows/test_wheels.yml
@@ -1,12 +1,12 @@
 name: test wheel builds
 
 on:
-  pull_request_review_comment:
-    types: [created, edited]
+  pull_request:
+    types: [labeled]
 
 jobs:
   macos_wheel:
-    if: contains(github.event.comment.body, "/wheels_all") || contains(github.event.comment.body, "/wheels_mac")
+    if: contains(github.event.label.name, 'cibuildwheel-test') || contains(github.event.label.name, 'cibuildwheel-mac-test')
     runs-on: macos-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


splinepy/spline.py
```diff
@@ -266,9 +266,64 @@ def _call_required_properties(spl, exclude="?"):
...
+    array: np.ndarray
+    key: str
+    coupled_key: str
+
+    Returns
+    -------
+    copied: bool
+      Whether copy was made
+    """
+    # is current value same is given array?
```
j042, line 308:
```suggestion
    # is current value same as given array?
```

j042 (commented):
/wheels_all

.github/workflows/test_wheels.yml
```diff
@@ -1,12 +1,12 @@
 name: test wheel builds
 
 on:
-  pull_request_review_comment:
-    types: [created, edited]
+  pull_request:
+    types: [labeled]
```
clemens-fricke, line 5: Did not know this is possible nice find.

splinepy/spline.py
```diff
@@ -270,9 +270,76 @@ def _call_required_properties(spl, exclude="?"):
...
+
+def _safe_array_copy(spl, array, key, coupled_key):
+    """
+    In splinepy we make safe copies. For control_points and weights,
+    to support non-contiguous / homogeneous control points in backend,
+    we have a helper class called `PhysicalSpaceArray`. This syncs
+    numpy array and backend's control points through `ControlPointPointers`.
+
+    This function checks a non-copy criteria: self assignment.
+    This tends to happen during inplace operationrs.
```
clemens-fricke, line 295:
```suggestion
    This tends to happen during inplace operations.
```

clemens-fricke (approved):
From what I can see, this seems to be fine. 

Did not fully understand everything, but as I said it seems to be fine.

.github/workflows/test_wheels.yml
j042: :D

clemens-fricke (approved):

jzwar (approved):
as discussed : LGTM, thanks

PR #171, opened 2023-06-21
gustaf - Flip dependency

j042:
# Overview
Moves gustaf-spline functionalities to here, splinepy. Avoid some duplicating implementations and draws a cleaner separation. related to tataratat/gustaf#131

## Addressed issues
*  [this discussion](https://github.com/tataratat/gustaf/discussions/129)

## Showcase
A short / one-liner example to highlight the (new) feature
```
import gustaf as gus
import splinepy

splines.show()

gus.show(spline1, spline2)
```

## Checklists
* [ ] Documentations are up-to-date.
* [ ] Added example(s)
* [ ] Added test(s)

PR #311, opened 2023-11-17
remove _backend layer - match gustaf's #171

### LINKED ISSUE #323: Adapt to incomming changes to gustaf
Gustaf is removing the backend option (tataratat/gustaf#171) in most places and will from now on only use `vedo` as its visualization backend. This change necessitates some changes to splinepy's code base.

The needed changes are mainly in `splinepy/helpme/visualize.py`.

1. Remove references to SplineShowOption._backend
2. Make sure to make the Option definition into an iterable by having for single type options a comma after the element `(int)` -> `(int,)`

j042:
# Overview
This PR incorporates changes in tataratat/gustaf#171

and fixes #323

### DIFF
--- a/splinepy/helpme/visualize.py
+++ b/splinepy/helpme/visualize.py
@@ -15,7 +15,7 @@ _vedo_spline_common_options = (
         "knot_lw",
         "Line width of knots. Number of pixels. "
         "Applicable to para_dim > 1.",
-        (int,),
+        (float, float),
     ),
     _options.Option(
         "vedo",
@@ -70,7 +70,7 @@ _vedo_spline_common_options = (
         "vedo",
         "control_mesh_lw",
         "Line width of control mesh. Number of pixels",
-        (int,),
+        (float, int),
     ),
     _options.Option(
         "vedo",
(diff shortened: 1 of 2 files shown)


splinepy/helpme/visualize.py
```diff
@@ -15,7 +15,7 @@
         "knot_lw",
         "Line width of knots. Number of pixels. "
         "Applicable to para_dim > 1.",
-        (int),
+        (int,),
```
clemens-fricke, line 18: can also be a float, not so relevant if used in conjunction with the new vedo release but before you needed lw of 0.1 etc. I am not sure if we should add it now since it does not serve a useful purpose anymore. More just a find I made. (I have tested this just with he k3d backend in vedo so I am not even sure if this holds for the vtk backend.
```suggestion
        (int,float),
```

splinepy/helpme/visualize.py
```diff
@@ -70,7 +70,7 @@
         "vedo",
         "control_mesh_lw",
         "Line width of control mesh. Number of pixels",
-        (int),
+        (int,),
```
clemens-fricke, line 73: As above.

clemens-fricke (approved):
There might be even more of the kind I described above like for the control points. 

The changes are not needed. Even without them, I approve this PR. Nice work, thank you. This also enables the ipynb support to work (that is currently nearly done in gustaf) to work.

clemens-fricke (approved):
Thanks, great work. :D

PR #314, opened 2023-11-23
Update Documentation in HelpMe Modules

jzwar:
# Overview
Just some minor modifications to the documentation. Enables `help` calls on `spline.extract.<method>` and proper docs

## Checklists

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,11 @@
 splinepy welcomes and appreciates discussions, issues and pull requests!
 
 ## Quick start
+<<<<<<< Updated upstream
 Once the repo is forked, one possible starting point would be creating a new python environments, for example, using [conda](https://docs.conda.io/en/latest/miniconda.html) with `python=3.10`
+=======
+Once the repo is forked, one possible starting point would be creating a new python environments, for example, using [conda](https://docs.conda.io/en/latest/miniconda.html) with, for example, `python=3.9`
+>>>>>>> Stashed changes
 ```bash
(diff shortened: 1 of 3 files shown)


splinepy/helpme/extract.py
```diff
@@ -585,3 +608,15 @@ def spline(self, splitting_plane=None, interval=None):
             return spline_copy
         else:
             return spline(self._spline, splitting_plane, interval)
+
+
+# Use function docstrings in Extractor functions
+Extractor.edges.__doc__ = edges.__doc__
```
j042, line 614: I think you can integrate this inside the function

j042 (approved):
Thanks! Can you add a note about how to document helper's functions in `CONTRIBUTING.md` ?

splinepy/helpme/extract.py
jzwar: setting it within a function is not possible because the lines are not evaluated. within the class body also does not work, because you can not access these names as class variables. This is why I put them outside

j042:
```python
def c():
    """
    I am c
    """
    pass

class A:
    def b():
        pass

    b.__doc__ = c.__doc__

print(c.__doc__)
print(A.b.__doc__)
```

CONTRIBUTING.md
```diff
@@ -2,7 +2,11 @@
 splinepy welcomes and appreciates discussions, issues and pull requests!
 
 ## Quick start
+<<<<<<< Updated upstream
```
j042, line 5: clean up

Issue #291, opened 2023-11-07
Plotting functions on splines is too complicated


jzwar:
Plotting a function on a spline is very complicated at the moment, as every time the `resolution` and `on` case need to be dealt with. It is not possible to just plot the same function on a Multipatch. That makes it very hard for new users to understand.

I think it would be nice to have a layout like this

```python
def foo(spline, queries):
  # Function in physical space
  x = spline.evaluate(queries)
  return x*x

def bar(spline, queries):
  # Function defined in parametric space
  return queries*queries
```

j042:
#318

PR #318, opened 2023-12-06
Easier spline data plotting with callbacks

j042:
# Overview
This PR wishes to make plotting with callbacks easier. It removes requirement of defining the behavior with the `resolutions` argument.

## Addressed issues
*  #291 

## Q
I wanted to add this query helper to `helpme.create`, but @jzwar mentioned that `helpme.create` is quite exclusively for splines. Would you prefer to have a `helpme.make` to add some non-spline based makings?

### DIFF
--- a/examples/run_all_examples.py
+++ b/examples/run_all_examples.py
@@ -1,14 +1,19 @@
 """Call all *.py files in the working directory.
 """
 import glob
+import itertools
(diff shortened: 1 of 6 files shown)


jzwar (approved):
LGTM, thanks. Especially for adding all this documentation

PR #332, opened 2024-01-02
take max for resolution

j042:
# Overview
This PR enables extract.edges non-integer resolution input. This was previously an error, but now we debug log it and take max.


jzwar (approved):
Shouldn't this rather be a warning? Might be confusing if you change the resolution and everything stays the same 🤔 maybe we should use a separate keyword "resolutions". That would be my preference

PR #334, opened 2024-01-02
split proximity hpp / cpp

j042:
# Overview
splits proximity.hpp to headers and source. Should take less time to compile

### DIFF
--- a/splinepy/spline.py
+++ b/splinepy/spline.py
@@ -1457,13 +1457,17 @@ class Spline(_SplinepyBase, _core.PySpline):
 
         queries = _utils.data.enforce_contiguous(queries, dtype="float64")
 
+        # set small tolerance.
+        if tolerance is None and _settings.TOLERANCE > 1.0e-18:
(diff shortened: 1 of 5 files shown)


src/proximity/proximity.cpp
```diff
@@ -0,0 +1,288 @@
+#include <napf.hpp>
```
j042, line 1: rm

splinepy/spline.py
```diff
@@ -1457,13 +1457,17 @@ def proximities(
 
         queries = _utils.data.enforce_contiguous(queries, dtype="float64")
 
+        # set small tolerance.
+        if tolerance is None and _settings.TOLERANCE > 1.0e-18:
```
jzwar, line 1461: Interesting

jzwar (approved):
LGTM

splinepy/spline.py
j042: yes

PR #337, opened 2024-01-11
accept any number of order-queries for derivatives

j042:
# Overview
Implements #248

## Addressed issues
*  #248

## Showcase
```python
import splinepy
import numpy as np

q = np.random.random((10000, spline.para_dim))

assert np.allclose(
    spline.jacobian(q),
    spline.derivative(q, np.eye(spline.para_dim)).transpose(0, 2, 1)
)
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar (approved):
Thanks for Adding the tests, LGTM

PR #340, opened 2024-01-24
Fix: docs now require single core build, takes way to long but otherw…

clemens-fricke:
# Overview
Description

## Addressed issues
*  Docs build failed

## Showcase
Remove parallel build of docs.
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


j042 (approved):
Awesome, thanks for a quick fix!

PR #341, opened 2024-01-26
fix #259

### LINKED ISSUE #259: io.cats keywords expects `x` `y` `z` `t`
instead of `x0`,` x1`, .... It seems to be hardcoded for basic control points of 1-4 Dim splines.

j042:
Give control point var names in a way that up to 4 dim, it would be `x`, `y`, `z`, `t`, and afterwards, `x{dim}`.


jzwar (approved):
Thank you! I still think it's a bad format, but ok... LGTM

clemens-fricke (approved):
Seems to be fine. LGTM. Thanks.

PR #345, opened 2024-02-02
remove deprecated kwarg

j042:
# Overview
compatibility bug fix

## Addressed issues
*  deprecated kwarg

@clemens-fricke can you add ffd test here again?


j042:
@clemens-fricke I relocated your commit as it will fail all the tests in CI. It is in `fix-ffd-add-test` branch for now

PR #350, opened 2024-02-05
helpers use weakref

### LINKED ISSUE #327: use `weakref` to save "helpee"
Helpers classes keep pure reference of the "helpee" object. This results in a circular/cyclic reference. Python is quite good at finding those and release them, but we should avoid it by using `weakref`

j042:
# Overview
to avoid circular reference, we keep weakref of the helpee within helpers. See tataratat/gustaf#185.
Turns out, keeping everything weakref doesn't allow "chained"-one-line uses. So, we will just create helpers that don't have states on the fly. Ones that have states will use weakref - they are okay, since we won't use them in "chained"-one-lines.

## Addressed issues
*  fixed #327

### DIFF
--- a/splinepy/helpme/check.py
+++ b/splinepy/helpme/check.py
@@ -1,4 +1,5 @@
 import numpy as _np
+from gustaf.helpers._base import HelperBase as _HelperBase
 
 
 def valid_queries(spline, queries):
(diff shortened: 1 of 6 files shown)


jzwar (approved):
LGTM, thanks for fixing copy-paste mistakes and typos, too!

splinepy/helpme/check.py
```diff
@@ -53,7 +54,7 @@ def valid_queries(spline, queries):
     return True
 
 
-class Checker:
+class Checker(_HelperBase):
```
j042, line 57: no need for helperbase

PR #351, opened 2024-02-05
update cibuildwheels

j042:
# Overview
fixes windows powershell issue


clemens-fricke (approved):
Can not check due to no access to windows. If the windows builds and tests run it should be fine. Especially since these changes are only in the workflow files and changes are only in version numbers of the used tools.

j042:
yes - the label triggers window build test. see [here](https://github.com/tataratat/splinepy/actions/runs/7785400180)

PR #356, opened 2024-02-05
Selective Field Output in G+smo

jzwar:
# Overview
Field output is not in the usual gismo standard and only works together with [pygadjoint](https://github.com/tataratat/pygadjoints/tree/main) so I decided to extend the export functionality a little more.

## Addressed issues
*  Macro spline optimization sometimes only applies to a subset of all macro-spline ctps

@j042 I need this for optimisations only. I can have it on a separate branch for all eternity if you prefer. Also relevant for @kkILSB

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/io/gismo.py
+++ b/splinepy/io/gismo.py
@@ -48,8 +48,8 @@ def _spline_to_ET(
     if fields_only and (len(multipatch.fields) == 0):
         return
 
-    if fields_only and fields_only is not None:
-        if not isinstance(fields_only, (list, np.ndarray)):
+    if fields_only and field_mask is not None:
+        if not isinstance(field_mask, (list, np.ndarray)):
             raise ValueError("field_mask must be list of integers")
         field_mask = np.unique(
             enforce_contiguous(field_mask, dtype=np.int64, asarray=True)
--- a/splinepy/io/gismo.py
+++ b/splinepy/io/gismo.py
@@ -17,6 +17,7 @@ def _spline_to_ET(
     index_offset,
     fields_only=False,
     as_base64=False,
+    field_mask=True,
 ):
     """
     Write spline data to xml element in gismo format
@@ -35,6 +36,8 @@ def _spline_to_ET(
       If set, exports only the fields associated to the multipatch data
     as_base64 : bool (False)
       If set, coordinates, weights and knot-vectors are exported in B64
+    field_mask : array-like
+      Selection of active fields to be exported into the fields file
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


j042:
is it also relevant for @kkCATS?

splinepy/io/gismo.py
```diff
@@ -17,6 +17,7 @@ def _spline_to_ET(
     index_offset,
     fields_only=False,
     as_base64=False,
+    field_mask=True,
```
j042, line 20: why is this True if you expect array-like input?

splinepy/io/gismo.py
```diff
@@ -45,6 +48,22 @@ def _spline_to_ET(
...
+    if fields_only and field_mask is not None:
+        if not isinstance(field_mask, (list, _np.ndarray)):
+            raise ValueError("field_mask must be list of integers")
+        field_mask = _np.unique(
+            _enforce_contiguous(field_mask, dtype=_np.int64, asarray=True)
+        )
+
+        # Check if range is valid
+        if (field_mask.min() < 0) or (
+            field_mask.max() >= len(multipatch.fields)
```
j042, line 60:
```suggestion
        if (min(field_mask) < 0) or (
            max(field_mask) >= len(multipatch.fields)
```
this should be applicable for any array-like obj

splinepy/io/gismo.py
```diff
@@ -45,6 +48,22 @@ def _spline_to_ET(
     if fields_only and (len(multipatch.fields) == 0):
         return
 
+    if fields_only and field_mask is not None:
+        if not isinstance(field_mask, (list, _np.ndarray)):
+            raise ValueError("field_mask must be list of integers")
+        field_mask = _np.unique(
+            _enforce_contiguous(field_mask, dtype=_np.int64, asarray=True)
```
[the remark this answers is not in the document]
jzwar: `field_mask` is sometimes list -> need array, so yes

splinepy/io/gismo.py
jzwar, line 20:
```suggestion
    field_mask=None,
```
Good point, thanks @j042

splinepy/io/gismo.py
jzwar: is always numpy

j042: ok

splinepy/io/gismo.py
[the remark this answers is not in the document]
j042: ok

j042 (approved):
Thanks! It'd be sad to have a feature stay in a separate branch alone and for eternity

PR #361, opened 2024-02-12
remove outdated module

j042:
goodbye `load.py`


clemens-fricke:
I am not sure if the functionality is entirely covered by the `io/default` module. Would you like to remove this since the default module now exists? In that case, we should look at what this one is doing, which is not currently covered by `io/default`.

j042:
No, this module was just a quick fix / use module for back then. I thought I removed it, but today I realized that I didn't

PR #364, opened 2024-02-13
update install components

j042:
# Overview
Installing all the targets makes a lot of things easy, but the size of wheel is getting out of hand due to explicit instantiation. So this PR removes all-inclusive target exports. 

## Addressed issues
*  huge wheels
* pypi storage limit


jzwar:
Why did you remove the format runner?

j042:
because we have pre-commit.ci and having one less job gets another build job to start earlier on busy days

Issue #352, opened 2024-02-05
numpy DeprecationWarning in helpme/fit.py


j042:
```
splinepy/splinepy/helpme/fit.py:130: DeprecationWarning:
Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future.
Ensure you extract a single element from your array before performing this operation.
(Deprecated NumPy 1.25.)
    knot_vector[j + degree] = (1 - alpha) * u_k[i - 1] + alpha * u_k[i]
```

j042:
@FSchwar

j042:
#367

PR #367, opened 2024-02-14
Small bugfix

jzwar:
# Overview
`numpy` throws deprecation warning - now no longer

### DIFF
--- a/splinepy/helpme/fit.py
+++ b/splinepy/helpme/fit.py
@@ -123,13 +123,16 @@ def compute_knot_vector(degree, n_control_points, u_k, n_fitting_points):
         # NURBS Book eq. (9.68) (n_fitting_points = m + 1,
         # n_control_points = n + 1)
         d = (n_fitting_points) / (n_control_points - degree)
+
+        u_k_flat = u_k.flat
+
         # NURBS Book eq. (9.69)
         for j in range(1, n_control_points - degree):
             i = int(j * d)
             alpha = (j * d) - i
-            knot_vector[j + degree] = (1 - alpha) * u_k[
-                i - 1, 0
-            ] + alpha * u_k[i, 0]
+            knot_vector[j + degree] = (1 - alpha) * u_k_flat[
+                i - 1
+            ] + alpha * u_k_flat[i]
     return knot_vector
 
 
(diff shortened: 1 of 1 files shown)


splinepy/helpme/fit.py
```diff
@@ -127,7 +127,9 @@ def compute_knot_vector(degree, n_control_points, u_k, n_fitting_points):
         for j in range(1, n_control_points - degree):
             i = int(j * d)
             alpha = (j * d) - i
-            knot_vector[j + degree] = (1 - alpha) * u_k[i - 1] + alpha * u_k[i]
+            knot_vector[j + degree] = (1 - alpha) * u_k[
+                i - 1, 0
+            ] + alpha * u_k[i, 0]
```
clemens-fricke, line 132: Are we sure that in all cases there is always just one element in the array? Otherwise, we would need to be more careful in how we change this line.

j042:
#352

splinepy/helpme/fit.py
j042: Why is u_k 2D? any special reason for that?

FSchwar: `u_k` is only 2D (n x 2) in the `surface`-fitting method and 1D (n x 1) in `curve`-fitting.
`compute_knot_vector` is then called with `u_k` in `curve` and `curve` is called with `u_k = u_k[0]` / `u_k = u_k[1]` for each dimension in `surface`.

Maybe we have to change the call of `curve` in the `surface` module to `u_k = u_k[:, 0]` /`u_k = u_k[:, 1]` instead of `u_k = u_k[0]` / `u_k = u_k[1]` because this seems unclear?

But in general, `u_k` should only be 1D in `compute_knot_vectors`, that's why I didn't add a second index.

jzwar: even nx1 is a 2D array. Alternatively we can `ravel()` to accommodate both cases

j042: yeah, at least within this function, you always expect `(n, 1)` arrays, so `ravel()`at the beginning (after shape check) should be fine

j042: nice flat

Issue #366, opened 2024-02-14
Pictures in documentation


jzwar:
The pictures are only shown in the local build, but somehow the online build is different:

[here](https://tataratat.github.io/splinepy/_generated/splinepy.io.svg.export.html#splinepy.io.svg.export) should be a picture, it should look like this:
![image](https://github.com/tataratat/splinepy/assets/63095162/4fae9026-e8d1-4645-bd25-92ff8f3df771)

@clemens-fricke, I know I messed this up, but could you help me fix this?

j042:
#368

PR #368, opened 2024-02-14
Fix: docs

clemens-fricke:
# Overview
Should now work in external docs.

## Addressed issues
*  #366 

## Restriction
I have copied the file to an exported folder, this has to be updated manually if you want to change it later on.

## Checklists


j042 (approved):
Fair fix, I guess. Is there a way to support what's done before?

j042:
Thanks :D

PR #369, opened 2024-02-14
Fix #360

### LINKED ISSUE #360: inplace knot vector update
noticed from #297:
```python
spline.knot_vectors[0] = spline.knot_vectors[0].numpy() * 2000
```
may not result in desired knot vector

j042:
# Overview
Mainly fixes #360 by exposing and wrapping `ParameterSpace`. This reduces some runtime bookkeeping complexities and avoids such situations in #360 as a whole.
It also:
- support unclamped knots. at least extends up to proper evaluation. knot / degree manipulation is not supported yet (TODO - plotting knots raises error, as some unique_knots are located in forbidden areas)
- fixes minor issues in knot vectors
- bsplinelib cleanup

## Addressed issues
*  fixes #360
* issues with unclamped spline evaluation at last valid knot
* invalid inplace update of first and last knots

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/include/splinepy/py/py_spline.hpp
+++ b/include/splinepy/py/py_spline.hpp
@@ -144,6 +144,12 @@ public:
   /// For non-rational splines, only the former.
   py::tuple CoordinatePointers();
 
+  /// @brief Returns ParameterSpace meant to be
+  /// called library internally to prepare @property
+  /// @return
+  std::shared_ptr<bsplinelib::parameter_spaces::ParameterSpaceBase>
+  ParameterSpace();
+
   /// Returns knot vector of given dimension. meant to be
   /// called library internally to prepare @property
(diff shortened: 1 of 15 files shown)


jzwar (commented):
Not sure if this is the way to move forward. Allowing unclamped knotvectors opens a whole class of New problems. We need to rethink Bézier extraction, knot insertion, splitting, every operation that requires indices in the knot vector,  e.g. fitting and also knot vector bounds.

j042:
We have been supporting unclamped knot vectors. This one just fixes evaluation at "last valid" knot. This doesn't change any other functionality. If you want to be safe, we can add some guards for check knot vectors beforehand in python side

jzwar (approved):
Thanks for the explanation. Let's add the warning and we're good to go. LGTM

PR #375, opened 2024-02-28
revert default only-python-install

j042:
# Overview
Reverting part of #364, where we only install `PythonModule`. For local builds, it is very convenient to have them installed (at least for my use case). For wheel builds and distribution, we keep this to reduce wheel size. We can perhaps set a specific version of python and add all the targets only for that one.

## Addressed issues
*  Need all components for local use


jzwar (approved):
LGTM, 👌

PR #376, opened 2024-02-28
add __array_wrap__

j042:
# Overview
using numpy's ufunc, PhysicalSpaceArray won't sync afterward. Now they do.

## Addressed issues
*  `np.add(spline.cps, values, out=spline.cps)` won't sync core control points

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


clemens-fricke (approved):
LGTM, if tests and checks run through. Thanks.

jzwar (approved):
Nice catch!! I wouldnt even have known where to start

PR #377, opened 2024-02-28
mute local svg test

j042:
# Overview
base64 test always gives error due to inconsistent versions and compression. We just throw warning for that instead of error.

## Addressed issues
*  local test always fails, but results have no qualitative difference

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


jzwar:
image

🥳

jzwar (approved):

jzwar (commented):
That's one way, I would suggest to delete the test entirely and just check that it doesn't fail during the export. Maybe we can just export the file and write a message asking the developer to open the file in a browser

Issue #379, opened 2024-02-28
Error in example show_microstructures - DoubleLattice


clemens-fricke:
The example `show_microstructure` produces an error. Relevant lines beginning 174ff. 

I get the following error:
```
2024-02-28 14:19:27,280 splinepy [WARNING] DoubleLattice now expects 2 values
New call [[0.2034375]]
2024-02-28 14:19:27,281 splinepy [WARNING] DoubleLattice now expects 2 values
New call None
Traceback (most recent call last):
  File "/home/clemensfricke/Documents/git/splinepy/examples/show_microstructures.py", line 190, in <module>
    generator.show(
  File "/home/clemensfricke/Documents/git/splinepy/splinepy/microstructure/microstructure.py", line 688, in show
    microtile = self.microtile.create_tile(**kwargs)
  File "/home/clemensfricke/Documents/git/splinepy/splinepy/microstructure/tiles/double_lattice.py", line 74, in create_tile
    self.check_params(parameters)
  File "/home/clemensfricke/Documents/git/splinepy/splinepy/microstructure/tiles/tile_base.py", line 96, in check_params
    raise TypeError(
TypeError: Mismatch in parameter size, expected 1 x 1 but is 1 x 2
```

I have added some new additional information to the output. For example, whenever a new tile is created, I print out `New call {paramters}` and return the expected and present dimensions of the parameter array in the `check_paramters` function.

I guess that at some point, the tile is first called with only one parameter but later called with no parameters. This generates default parameters with two parameters, producing an error in the `check_parameter` function since there is still an object variable that is used to enable one parameter call but inhibits two parameter calls.

Could you check this out, @mkofler96?

PR #380, opened 2024-02-28
fixed double lattice parametrization

mkofler96:
# Overview
examples/show_microstructures.py works now

## Addressed issues
#379 

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


splinepy/microstructure/tiles/double_lattice.py
```diff
@@ -57,19 +57,19 @@ def create_tile(
...
             and _np.all(parameters < 0.5 / (1 + _np.sqrt(2)))
         ):
             raise ValueError(
                 "Parameters must be between 0.01 and 0.5/(1+sqrt(2))=0.207"
             )
+        # Maintain backwards compatibility
+        if parameters.shape[1] == 1:
+            _warning("DoubleLattice now expects 2 values")
+            index_second_value = 0
+            self._n_info_per_eval_point = 1
```
jzwar, line 72: Please delete this section! It's not that it would not work, it's just avoidable

splinepy/microstructure/tiles/double_lattice.py
```diff
@@ -57,19 +57,19 @@ def create_tile(
         # set to default if nothing is given
         if parameters is None:
             self._logd("Tile request is not parametrized, setting default 0.2")
-            parameters = _np.ones((1, 2)) * 0.1
-        # Maintain backwards compatibility
-        elif parameters.shape[1] == 1:
-            _warning("DoubleLattice now expects 2 values")
-            index_second_value = 0
-            self._n_info_per_eval_point = 1
+            parameters = _np.ones((1, self._n_info_per_eval_point)) * 0.1
```
jzwar, line 60:
```suggestion
            parameters = _np.ones((1, self._n_info_per_eval_point)) * 0.1
                    # Maintain backwards compatibility
         elif parameters.shape[1] == 1:
             _warning("DoubleLattice now expects 2 values")
             self._index_second_value = 0
             self._n_info_per_eval_point = 1
```

It's cleaner if the index is a member variable. Another option is to discard the variable completely and replace the calls of index using `n_info-1`, but that is very hard to read. You could also throw a deprecation warning.

jzwar (changes requested):

splinepy/microstructure/tiles/double_lattice.py
mkofler96: I don't know how to avoid this 🥲

splinepy/microstructure/tiles/double_lattice.py
mkofler96: Member variable definitely makes sense. I also thought about the n_info - 1 and I also think it makes it harder to read.

j042 (commented):
I don't like those warnings. You should either support only 2 parameters or document behaviors for 1 parameter and do that

mkofler96:
> I don't like those warnings. You should either support only 2 parameters or document behaviors for 1 parameter and do that

I'm fine with both versions, slightly favoring supporting only 2 parameters. What do you think @jzwar?

jzwar:
> > I don't like those warnings. You should either support only 2 parameters or document behaviors for 1 parameter and do that
> 
> I'm fine with both versions, slightly favoring supporting only 2 parameters. What do you think @jzwar?

As discussed in person. 2 parameters is fine with me, though we should rather focus on 4 parameters, as 2 is not really beneficial in structural use-cases

j042:
> > > I don't like those warnings. You should either support only 2 parameters or document behaviors for 1 parameter and do that
> > 
> > 
> > I'm fine with both versions, slightly favoring supporting only 2 parameters. What do you think @jzwar?
> 
> As discussed in person. 2 parameters is fine with me, though we should rather focus on 4 parameters, as 2 is not really beneficial in structural use-cases

@mkofler96 can you make it to require 2 params for now (and 4 params later ;) ) ?

mkofler96:
> > > > I don't like those warnings. You should either support only 2 parameters or document behaviors for 1 parameter and do that
> > > 
> > > 
> > > I'm fine with both versions, slightly favoring supporting only 2 parameters. What do you think @jzwar?
> > 
> > 
> > As discussed in person. 2 parameters is fine with me, though we should rather focus on 4 parameters, as 2 is not really beneficial in structural use-cases
> 
> @mkofler96 can you make it to require 2 params for now (and 4 params later ;) ) ?

did it for 2 parameters for now. The 4 parameters-version is a bit more complicated (see https://github.com/mkofler96/splinepy/tree/ft-quad-lattice) and we are not completely happy yet with the parametrization, since it is kind of ugly

splinepy/microstructure/tiles/double_lattice.py
```diff
@@ -58,18 +56,18 @@ def create_tile(
...
             _np.all(parameters > 0)
             and _np.all(parameters < 0.5 / (1 + _np.sqrt(2)))
         ):
             raise ValueError(
                 "Parameters must be between 0.01 and 0.5/(1+sqrt(2))=0.207"
             )
+        # raise an error if only one parameter is given,
+        # which was the case in earlier versions
+        if parameters.shape[1] == 1:
+            raise ValueError("DoubleLattice now expects 2 values")
```
jzwar, line 69: Please delete. We don't need deprecation warnings, we are beta

jzwar (approved):

clemens-fricke:
Superceeded by #400.

PR #381, opened 2024-02-29
*WIP stubgen - does not work because new functions are not added when rerunnning stubgen

mkofler96:
# Overview
There is the possibility to automatically generate .pyi files for type hinting by [mypy](https://mypy.readthedocs.io/en/stable/stubgen.html), which can by installed by `pip install mypy` and run by `stubgen splinepy`. The auto-generated .pyi files look like this:

```Python
class BSpline(BSplineBase):
    def __init__(self, degrees: Incomplete | None = None, knot_vectors: Incomplete | None = None, control_points: Incomplete | None = None, spline: Incomplete | None = None) -> None: ...
    @property
    def bspline(self): ...
    @property
    def nurbs(self): ...
```

and need to be translated to somethings like this by using e.g. `import numpy.typing as _npt`

```Python
class BSpline(BSplineBase):
    def __init__(self, degrees: _npt.ArrayLike | None = None, knot_vectors: _npt.ArrayLike | None = None, control_points: _npt.ArrayLike | None = None, spline: Incomplete | None = None) -> None: ...
    @property
    def bspline(self): ...
    @property
    def nurbs(self): ...
```

The stub generation is not useful for final use, but it still provides at least the basic layout of the .pyi files.

What do you think? @j042 @jzwar @clemens-fricke


clemens-fricke:
Looks ok will have to see how much better the autocomplete will work with these added. 

After a quick look, a lot of the files are superfluous since they are not able to add typing information examples most __init__ files.

mkofler96:
> Looks ok will have to see how much better the autocomplete will work with these added.
> 
> After a quick look, a lot of the files are superfluous since they are not able to add typing information examples most **init** files.

I think there is still manual work needed to overwrite the `Incomplete`, but at least one does not have to write each entire .pyi file. 
Also, what I saw is that if you run the `stubgen splinepy` command when there are already .pyi files present, the types are not overwritten to `Incomplete`, but the correct type is kept.

PR #382, opened 2024-02-29
*WIP* Inline type hints

clemens-fricke:
# !!! Currently no type hints !!!
I have started with the example overall similar to https://github.com/tataratat/gustaf/pull/197 to enable type hint tracing also in the examples.
*Example tracing with `run_all_examples` is not possible :(*

Also, I wanted to enable us to run the examples in CI/CD to also check them.

# Overview
Similar to #381 but not in stub-files but with inline type hinting.

Using https://github.com/dropbox/pyannotate to trace tests to automatically detect the types enables us to add type hints quicker.

## Addressed issues
* Type hints are available
* Autocomplete in IDE is better
* During further development adding type hints is easier than using stub-files
* Secondary benefit, check which functions are called during tests and examples

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

# Dependencies
* https://github.com/tataratat/gustaf/pull/197
  -> Due to using the gustaf setting `DEFAULT_OFFSCREEN`

### DIFF
--- a/examples/basis_and_support.py
+++ b/examples/basis_and_support.py
@@ -2,7 +2,8 @@ import numpy as np
 
 import splinepy
 
-if __name__ == "__main__":
+
+def example():
     kv = [[0, 0, 0, 0.5, 1, 1, 1], [0, 0, 0, 1, 1, 1], [0, 0, 1, 1]]
     cp1 = np.array(
         [
@@ -57,3 +58,7 @@ if __name__ == "__main__":
         *b.basis_and_support(q), b.control_points.shape[0], as_array=True
     )
     print(basis_function_matrix)
+
+
+if __name__ == "__main__":
+    example()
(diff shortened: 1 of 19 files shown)


clemens-fricke:
I have now tried to add the inline annotations traced from just the tests (using `pytest-annotate` pytest plugin). This is not as trivial as I had hoped. Basically everything exploded and after fixing some format issues, I have a lot of circular dependency errors. I think it might be good if we look into this during the upcoming HackDayz.

j042:
> I have now tried to add the inline annotations traced from just the tests (using `pytest-annotate` pytest plugin). This is not as trivial as I had hoped. Basically everything exploded and after fixing some format issues, I have a lot of circular dependency errors. I think it might be good if we look into this during the upcoming HackDayz.

Can you post some circular dependency errors?

j042:
Closing this for now. Feel free to reopen

PR #383, opened 2024-03-01
Bugfix to compute G1 correctly

jzwar:
We were just lucky that it worked so far. I think this should solve the issue

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/src/py/py_multipatch.cpp
+++ b/src/py/py_multipatch.cpp
@@ -814,8 +814,19 @@ int AddBoundariesFromContinuity(const py::list& boundary_splines,
   int* global_interfaces_ptr =
       static_cast<int*>(global_interfaces.request().ptr);
 
+  // This function is currently only implemented for 2D and 3D applications. For
+  // higher dimensions, please revise the implementation of the generalized
+  // cross product as can be found here:
+  // https://math.stackexchange.com/questions/752774/generalized-cross-product
+  if ((dim_ < 1) || (dim_ > 3)) {
+    splinepy::utils::PrintAndThrowError(
+        "G1-based boundary computation is only supported for 2D and 3D "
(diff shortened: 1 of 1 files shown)


j042 (approved):
Did it work so far since all the multipatches were well aligned?
So with the normals, you don't have to worry about having a correct partial derivative - is this the fix?

jzwar:
So far we did not check G1, but something stronger. As we required, that adjacent patches hat same normals on the surface

> Did it work so far since all the multipatches were well aligned?
> So with the normals, you don't have to worry about having a correct partial derivative - is this the fix?

It's a weaker condition

j042:
For para_dim==1, this is practically the same, and para_dim==2, it is a different behavior, correct?

j042:
Ok, after discussion, we conclude that this is G1 in a broad sense. Title was just confusing, as it wasn't wrong before

src/py/py_multipatch.cpp
```diff
@@ -833,13 +844,12 @@ int AddBoundariesFromContinuity(const py::list& boundary_splines,
...
-                               const int& patch_id,
-                               const int& face_id) -> std::vector<double> {
+  auto normal_vector = [&cpp_spline_list, &para_dim_, &dim_](
+                           const int& patch_id,
+                           const int& face_id) -> std::vector<double> {
     // init return value (are default initialized to 0)
     std::vector<double> para_coord(para_dim_), bounds(2 * para_dim_),
-        tangential_vector(dim_);
-    std::vector<int> orders(para_dim_);
+        normal_vector(dim_), jacobian(dim_ * para_dim_);
```
j042, line 852: Based on no warning/error, I guess this doesn't necessarily count as "over shadowing" variable?

Issue #385, opened 2024-03-01
Issue with Matrix solve


jzwar:
There seems to be an issue with our implementation of Gausselimination with partial pivoting. I added this block to `Solve()` in `include/splinepy/utils/array.hpp` 

```cpp
#ifndef NDEBUG
    double norm{};
    for (IndexType i{}; i < len; ++i) {
      double row_val{};
      for (IndexType j{}; j < len; ++j) {
        row_val += array_copy(i, j) * x[j];
      }
      row_val -= b_copy[i];
      norm += row_val * row_val;
    }
    if (std::sqrt(norm) > 1e-4) {
      splinepy::utils::PrintAndThrowError("Solving failed, error is ", norm);
    }
#endif
```
It throws a lot. There is no testing of this. So far I haven't figured out where the bug is yet.

jzwar:
Do you have any ideas @j042 ? Is my test wrong?

jzwar:
The error only occurs when pivoting is used

jzwar:
@j042 : Found the bug

j042:
Thank you!
#386

PR #386, opened 2024-03-01
Bugfix - Wrong indices for back substitution

jzwar:
The indices of the solution vector do not change as the pivoting only affects the rows of the matrix and the lhs, but not the solution (columns remain unchanged).

@j042, please verify

## Addressed issues
*  #385 

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/include/splinepy/utils/arrays.hpp
+++ b/include/splinepy/utils/arrays.hpp
@@ -548,6 +548,12 @@ public:
     assert(len == b.size());
     assert(len == x.size());
 
+#ifndef NDEBUG
+    // Create a copy of the linear system for debugging
+    const auto array_copy = array_;
+    const auto b_copy = b;
+#endif
+
     // partial pivoting and forward reduction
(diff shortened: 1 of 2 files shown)


jzwar:
Also :
I would like to delete the unused function: `GaussWithPivot` to avoid confusion

j042:
What a good catch, thanks! I agree with some clean up. Maybe we can add LAPACK based solve for debug or maybe use that as numpy always comes with lapack/blas

jzwar:
> What a good catch, thanks! I agree with some clean up. Maybe we can add LAPACK based solve for debug or maybe use that as numpy always comes with lapack/blas

@j042 If it is ok with you, I would just write an issue to push this to later. Imo, we can already merge this pr and wait with both clean up and bugfix for convergence issues for later...

j042:
Sure

j042 (approved):
Thanks!

PR #387, opened 2024-03-05
Mini-PR Linewidth for 1D splines

jzwar:
# Overview
As the title says.

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/helpme/visualize.py
+++ b/splinepy/helpme/visualize.py
@@ -78,6 +78,12 @@ _vedo_spline_common_options = (
         "Sampling resolution for spline.",
         (int, list, tuple, _np.ndarray),
     ),
+    _options.Option(
+        "vedo",
+        "lw",
+        "Line width 1D spline." "Applicable to para_dim > 1.",
+        (float,),
+    ),
 )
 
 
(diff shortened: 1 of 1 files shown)


clemens-fricke (approved):
Could you check if the name is the same as in Gustaf for edge line width. If it is perfect, but I am currently not sure.

splinepy/helpme/visualize.py
```diff
@@ -78,6 +78,12 @@
         "Sampling resolution for spline.",
         (int, list, tuple, _np.ndarray),
     ),
+    _options.Option(
+        "vedo",
+        "lw",
+        "Line width 1D spline." "Applicable to para_dim > 1.",
+        (float,),
+    ),
```
j042, line 86:
```suggestion
    # overwrites the option with a same name in vedo_common_options
    _options.Option(
        "vedo",
        "lw",
        "Line width 1D spline." "Applicable to para_dim > 1.",
        (float, int),
    ),
```

j042 (approved):
Thanks, always wanted those! only one suggestion

PR #388, opened 2024-03-05
Enforce Recomputation of Multipatch Interfaces

jzwar:
# Overview
As the title says.

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/include/splinepy/py/py_multipatch.hpp
+++ b/include/splinepy/py/py_multipatch.hpp
@@ -370,7 +370,9 @@ public:
   /// @brief setter and getter for interface info. runs array shape check for
   /// setter
   /// @param interfaces
-  py::array_t<int> Interfaces(const py::array_t<int>& interfaces);
+  /// @param recompute compute even if already available
+  py::array_t<int> Interfaces(const py::array_t<int>& interfaces,
+                              const bool recompute = false);
 
   /// @brief Gets IDs of boundary patch
   py::array_t<int> BoundaryPatchIds();
--- a/include/splinepy/py/py_multipatch.hpp
+++ b/include/splinepy/py/py_multipatch.hpp
@@ -367,12 +367,13 @@ public:
   /// @return
   py::array_t<double> SubPatchCenters();
 
-  /// @brief setter and getter for interface info. runs array shape check for
-  /// setter
+  /// @brief setter for interface info. runs array shape check
   /// @param interfaces
+  void SetInterfaces(const py::array_t<int>& interfaces);
+
+  /// @brief getter for interface info. recomputes if unavailable
   /// @param recompute compute even if already available
-  py::array_t<int> Interfaces(const py::array_t<int>& interfaces,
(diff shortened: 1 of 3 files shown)


j042:
When was this causing err?

jzwar:
> When was this causing err?

No error, but also no way to recompute once done. This just adds a new option.

j042:
> > When was this causing err?
> 
> No error, but also no way to recompute once done. This just adds a new option.

I think I've mentioned this back then: what about removing property/setter and use "vedo style":
```
mp.interface(new_inter)  # setter
interf = mp.interface()  # getter
```

jzwar:
> > > When was this causing err?
> > 
> > 
> > No error, but also no way to recompute once done. This just adds a new option.
> 
> I think I've mentioned this back then: what about removing property/setter and use "vedo style":
> 
> ```python
> mp.interface(new_inter)  # setter
> interf = mp.interface()  # getter
> recomp_interf = mp.interface(recompute=False) # getter
> ```

Naaaaa, I like the properties. In 99% of cases, you need no recomputation

j042:
> > > > When was this causing err?
> > > 
> > > 
> > > No error, but also no way to recompute once done. This just adds a new option.
> > 
> > 
> > I think I've mentioned this back then: what about removing property/setter and use "vedo style":
> > ```python
> > mp.interface(new_inter)  # setter
> > interf = mp.interface()  # getter
> > recomp_interf = mp.interface(recompute=False) # getter
> > ```
> 
> Naaaaa, I like the properties. In 99% of cases, you need no recomputation

1% is what matters

jzwar:
> > > > > When was this causing err?
> > > > 
> > > > 
> > > > No error, but also no way to recompute once done. This just adds a new option.
> > > 
> > > 
> > > I think I've mentioned this back then: what about removing property/setter and use "vedo style":
> > > ```python
> > > mp.interface(new_inter)  # setter
> > > interf = mp.interface()  # getter
> > > recomp_interf = mp.interface(recompute=False) # getter
> > > ```
> > 
> > 
> > Naaaaa, I like the properties. In 99% of cases, you need no recomputation
> 
> 1% is what matters

Let's not get political 😄 I like the properties

j042 (changes requested):
Can we go with setter and getter for cpp? I don't like that you have to give empty list to give a recompute flag (`interfaces = super().interfaces([], recompute)`)
```cpp
PyMultipatch::SetInterface(py::array_t<double>);

PyMultipatch::GetInterface(bool recompute);
```

jzwar:
> Can we go with setter and getter for cpp? I don't like that you have to give empty list to give a recompute flag (`interfaces = super().interfaces([], recompute)`)
> 
> ```c++
> PyMultipatch::SetInterface(py::array_t<double>);
> 
> PyMultipatch::GetInterface(bool recompute);
> ```

done

j042 (approved):
Great and Clear, thank you!

PR #393, opened 2024-03-10
Field dimension raise when field_dim != physical dimension

jzwar:
# Overview
See tests. Raise were not always reasonable

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


j042 (commented):
Can't you just use [this](https://github.com/tataratat/splinepy/blob/02734543cb59e85d35e04942b37323519027ff0c/src/py/py_multipatch.cpp#L89) instead?

j042:
> Can't you just use [this](https://github.com/tataratat/splinepy/blob/02734543cb59e85d35e04942b37323519027ff0c/src/py/py_multipatch.cpp#L89) instead?

I don't think so

j042 (approved):
After a closer look, it seems good. Thanks!

PR #396, opened 2024-03-12
Bug in spline extractions

jzwar:
There was a bug in spline extraction, as you need two different indexing schemes for "single-line" extraction and interval extraction (because in a slice, the last index is non-inclusive).

## Checklists
* [ ] Added example(s)
* [x] Added test(s)


clemens-fricke (approved):
Seems fine to me.

PR #401, opened 2024-03-20
More More More More SVG

jzwar:
# Overview
- clean-up : `write_png` is now separate function because, who knows when we might need it again
- `font-family` is no longer defaulted, because the standard font is nicer in most cases (allows for more flexibility)
- `control_points` had a bug -> it overwrote all other control mesh options, now we have the same behaviour as for `spp.show()`
- `arrow_data_scale` and `vmin/vmax` are now no longer fighting each other
- Most importantly:
**ADD A COLOR BAR**

## Checklists
* [x] Added example(s)
* [ ] Added test(s)


jzwar:
Added another feature, because:
- new flag `n_colors`

image

not as nice as

image

jzwar:
related to https://github.com/tataratat/gustaf/pull/201

j042 (approved):
Awesome and very useful! Thank you :)

PR #405, opened 2024-04-02
Bf gismo export

FSchwar:
# Overview
Fixed issues when exporting multipatches with `None` as field values. 

## Addressed issues
* multipatches with `None` in fields raised several errors

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/splinepy/io/gismo.py
+++ b/splinepy/io/gismo.py
@@ -109,10 +109,14 @@ def _spline_to_ET(
             [" ".join([str(xx) for xx in x]) for x in supports]
         )
 
+    id_count = -1
     for id, spline in enumerate(multipatch.patches):
         if fields_only:
             # Check supports
             support = supports[supports[:, 0] == id, 1]
+            if support.size == 0:
+                continue
+            id_count += 1
             coefs = _np.hstack(
                 [
                     multipatch.fields[j].patches[id].control_points
@@ -124,6 +128,7 @@ def _spline_to_ET(
                     [multipatch.fields[j][id].weights for j in support]
                 )
         else:
+            id_count = id
             coefs = spline.control_points
             if "weights" in spline.required_properties:
                 weights = spline.weights
@@ -150,7 +155,7 @@ def _spline_to_ET(
         spline_element = _ET.SubElement(
             root,
             "Geometry",
-            id=str(id + index_offset),
+            id=str(id_count + index_offset),
             type="Tensor" + type_name + str(spline.para_dim),
         )
 
@@ -209,6 +214,7 @@ def _spline_to_ET(
         )
         coords.text = _array_to_text(coefs, True, as_base64)
 
+    return id_count
 
 def export(
     fname,
@@ -299,9 +305,9 @@ def export(
     patch_range = _ET.SubElement(
         multipatch_element, "patches", type="id_range"
     )
-    patch_range.text = (
-        f"{index_offset} " f"{len(multipatch.patches) - 1 + index_offset}"
-    )
+    # patch_range.text = (
+    #     f"{index_offset} " f"{len(multipatch.patches) - 1 + index_offset}"
+    # )
 
     ###
     # Individual spline data
@@ -309,7 +315,8 @@ def export(
     # Export fields first, as all necessary information is already available
     if export_fields:
         field_xml = _copy.deepcopy(xml_data)
-        _spline_to_ET(
+
+        n_patches = _spline_to_ET(
             field_xml,
             multipatch,
             index_offset,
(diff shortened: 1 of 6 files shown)


splinepy/io/gismo.py
```diff
@@ -302,13 +309,45 @@ def export(
     patch_range = _ET.SubElement(
         multipatch_element, "patches", type="id_range"
     )
-    patch_range.text = (
-        f"{index_offset} " f"{len(multipatch.patches) - 1 + index_offset}"
-    )
+    # patch_range.text = (
+    #     f"{index_offset} " f"{len(multipatch.patches) - 1 + index_offset}"
+    # )
```
jzwar, line 314: Deprecated

splinepy/io/gismo.py
```diff
@@ -109,10 +109,14 @@ def _array_to_text(array, is_matrix, as_b64):
             [" ".join([str(xx) for xx in x]) for x in supports]
         )
 
+    id_count = -1
```
jzwar, line 112: Please add a comment why this is done

splinepy/io/gismo.py
```diff
@@ -302,13 +309,45 @@ def export(
...
+        with open(fname + ".fields.xml", "wb") as f:
+            f.write(file_content)
+
+    n_patches = _spline_to_ET(
+        xml_data,
+        multipatch,
+        index_offset,
+        as_base64=as_base64,
+    )
+    assert n_patches + 1 == len(multipatch.patches)
```
jzwar, line 346: if raise

jzwar (commented):
Left some remarks,

why do the tests fail. Please check!

FSchwar:
`test_gismo` fails in `test_gismo_export`at line 84 - where the lines of the `base_file` and the temporary exported file are compared.
I manually compared the files and it seems that only the order of the lines is off.

To be more precise, the definition of the `boundaryConditions` in the `base_file` is before the definition of `Geometry`, whereas in the temporary file (newly exported one), the definition of the boundary conditions is at the bottom of the xml file.

splinepy/io/gismo.py
```diff
@@ -302,13 +311,45 @@ def export(
...
+        field_xml.find("MultiPatch").find("patches").text = (
+            f"{index_offset} " f"{n_patches + index_offset}"
+        )
+        if int(_python_version.split(".")[1]) >= 9 and indent:
+            _ET.indent(field_xml)
+        file_content = _ET.tostring(field_xml)
+        with open(fname + ".fields.xml", "wb") as f:
+            f.write(file_content)
+
+    n_patches = _spline_to_ET(
```
jzwar, line 339: This Block of Code must stay where it was otherwise the order will change

splinepy/io/gismo.py
```diff
@@ -80,7 +80,7 @@ def _array_to_text(array, is_matrix, as_b64):
             field_mask = _np.arange(len(multipatch.fields))
         supports = _np.array(
             [
-                (i, j, 0)
+                (i, k, 0)
```
jzwar, line 83: I think we don't need the enumerate block here do we? J is unused

splinepy/io/gismo.py
FSchwar: It's successful now, thanks!

splinepy/io/gismo.py
```diff
@@ -109,10 +109,16 @@ def _array_to_text(array, is_matrix, as_b64):
...
+    # set to -1 since loop starts with id_count += 1, therefore the first
+    # patch starts with id = 0
+    id_count = -1
     for id, spline in enumerate(multipatch.patches):
         if fields_only:
             # Check supports
             support = supports[supports[:, 0] == id, 1]
+            if support.size == 0:
+                continue
+            id_count += 1
```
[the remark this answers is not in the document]
FSchwar: not needed (as discussed)

splinepy/io/gismo.py
FSchwar: was changed back again in a later commit for successful field mask

jzwar:
There is still one issue that is unsolved in opinion at the moment. I think the reader will break if the spline ids are not contiguous, meaning if we only use splines `[0,1,2,4,5]` it will most likely break. This is because the the `support` matrix is referring to the splines in their global configuration, whereas the splines themselves are simply counted. There is two possible solutions to this issue. Either we introduce another column in the support matrix, where the global ids are matched a local id, this would have to be updated [here](https://github.com/tataratat/pygadjoints/blob/34aa2f287a0b8b8e453a865548d112bb80128b75/src/pygadjoints/py_elasticity.hpp#L610C17-L610C24) and in the respective diffusion field. Second would not be a fix, but rather a warning, where the user is warned (in case a `id_count!=id`. 

Please implement the latter, and I will make sure to update whenever this becomes an issue. Just make an elaborate warning, whenever this problem pops up

PR #407, opened 2024-04-10
Ft extract basis

j042:
# Overview
Visualizing basis is easier now. `spline.extract.bases()` will create a spline that represents the basis function. This isn't the most efficient way, but fits right into our current visualization and helpers pipeline.
Some sort of automatic random coloring could be a nice add-on.

## Addressed issues
*  Visualizing basis was not easy.

## Showcase
```python
bases = spline.extract.bases()
spp.show(bases)

support123 = spline.extract.bases([1, 2, 3])
spp.show(support123)
```
1d
2d

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/examples/show_basis.py
+++ b/examples/show_basis.py
@@ -1,0 +1,81 @@
+import splinepy as spp
+
+if __name__ == "__main__":
+    # curve
+    # define degrees
+    ds1 = [2]
+
+    # define knot vectors
+    kvs1 = [[0.0, 0.0, 0.0, 1.0, 1.0, 1.0]]
+
+    # define control points
+    cps1 = [
+        [0.0, 0.0],
+        [-2.0, 4.0],
+        [1.0, 3.0],
+    ]
+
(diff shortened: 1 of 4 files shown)


jzwar:
Unrelated: we should bump version soon

j042:
> Nice idea! You could just add a line with `show_options["c"]=i` to automatically display in different colors.
> 
> The spline extraction also works for rationals so unless you don't want to use meshes or matplotlib, this is a pretty efficient way to visualize actually...

AND many thanks to `svg` export, we can export smooth bases, smooth!

good call with `"c"=i`

j042:
> Unrelated: we should bump version soon

yes, was waiting to finish a few things to bump it to `0.1`, but I think it's better to bump one with this or something.

Also wanna add a test for this. Maybe will add directly to pytest PR

j042:
@jzwar
e1
e2

PR #408, opened 2024-04-10
Ft cpp doc

j042:
# Overview
adding some cpp documentations. We use doxygen to create docs and just add link to the doxygen main page. Originally, I wanted to add docs for some classes with breathe, but it's clearer and cleaner this way.

## Addressed issues
*  no cpp docs

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


j042:
@markriegler could you help me get read of some doxygen warnings?

markriegler:
I can help

PR #410, opened 2024-04-15
fix pytest import issue

j042:
# Overview
Installing splinepy non-editable mode would cause import issues. Mainly due to the compiled module, as it is not located in local directory. As stated in [pytest doc](https://docs.pytest.org/en/stable/explanation/pythonpath.html), this PR takes `append` mode for pytest to prefer installed packaged.

## Addressed issues
*  failed local / CI test calls


jzwar (approved):
LGTM, thanks!

PR #411, opened 2024-04-15
remove getters in conftest

j042:
# Overview
replace `get_*` by hardcoding values.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,79 +18,6 @@ def np_rng():
     return np.random.default_rng()
 
 
-@pytest.fixture
-def get_2d_control_points_b_spline():
-    return [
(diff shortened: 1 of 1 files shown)


clemens-fricke (approved):
I thought they were used somewhere else as well. Do we not have a separate init test?

If we do not use them, thank you for finding the unnecessary function. 

LGTM.

j042:
> I thought they were used somewhere else as well. Do we not have a separate init test?
> 
> If we do not use them, thank you for finding the unnecessary function.
> 
> LGTM.

we do [here](https://github.com/tataratat/splinepy/blob/main/tests/test_spline_constructors.py). but we test it by creating some specific ones. these are there for some tests against reference values

PR #417, opened 2024-04-18
Readme updates

j042:
# Overview
Updated readme and documentations to give a better overview of this project. I've also included some chores / clean ups here so that examples in readme works as it is.
Changes are:
- more shortcuts. frequently used / main features are directly accessible at `splinepy.*`
- fitting clean up. tries to use less ifs. It seems that there's an issue with approximation
- microstructure tiles clean up. set some variables as class variable
- added module functions for tiles - easy access to tiles

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)


j042 (commented):
```python
    @property
    def microtile(self):
        """Microtile that is either a spline, a list of splines, or a class
        that provides a `create_tile` function."""
        if hasattr(self, "_microtile"):
            return self._microtile
        else:
            _log.info("microtile is empty. Please checkout splinepy.microstructure.tiles.show() for predefined tile collections!")
            return None
```

splinepy/helpme/fit.py
```diff
@@ -231,6 +223,76 @@ def solve_for_control_points(
...
+
+    # we need degree
+    if degree is None:
+        raise ValueError(
+            "Not enough input to determine degree. Please set degree."
+        )
+
+    # no n_control point -> same as query and this will be interpolation
+    if n_control_points is None:
+        n_control_points = n_query
```
danielwolff1, line 255: Does it make sense to inform the user here via a warning or a log message what value is chosen for `n_control_points` under the hood?

splinepy/microstructure/microstructure.py
```diff
@@ -135,6 +135,11 @@ def microtile(self):
         if hasattr(self, "_microtile"):
             return self._microtile
         else:
+            self._logi(
```
danielwolff1, line 138: I don't know what your convention is here, but shouldn't this be a warning at least, as something was requested which seems to not be defined?

splinepy/microstructure/tiles/ellips_v_oid.py
danielwolff1, line 1: Can it be that there went something wrong with the additional underscore in the file name here, or is this desired?

splinepy/microstructure/tiles/snappy.py
```diff
@@ -5,14 +5,14 @@
...
-        self._para_dim = 2
-        # Dummy point
-        self._evaluation_points = _np.array([[0.5, 0.5]])
-        self._n_info_per_eval_point = 1
+    """Snap-through tile consisting of a thin truss and a thick truss that
+    collide into each other"""
+
+    _dim = 2
+    _para_dim = 2
+    # dummpy values - not used
```
danielwolff1, line 13:
```suggestion
    # dummy values - not used
```

splinepy/microstructure/tiles/tile_base.py
```diff
@@ -8,18 +8,33 @@ class TileBase(_SplinepyBase):
...
-        Init Values to None
-        """
-        self._dim = None
-        self._evaluation_points = None
-        self._n_info_per_eval_point = None
+        if type(self) is TileBase:
+            raise TypeError(
+                "Can't initialize a TileBase class! Please use one of the "
+                f"derived classes: {TileBase.__subclasses__()}"
+            )
```
danielwolff1, line 21: I assume this code is supposed to prevent an instantiation of an instance of type `TileBase`. Is there a reason why you decided to go this way instead of inheriting from `abc.ABC`?

danielwolff1 (approved):
Looking great, I really really like the modified documentation!

I just had a few minor remarks / questions but generally I approve of the changes made here and you can proceed with the merge! 🚀

splinepy/microstructure/tiles/ellips_v_oid.py
j042: I asked the same question to @jzwar and I got a reply saying that I should have a better sense of humor

splinepy/microstructure/microstructure.py
j042: `i`nfo, `d`ebug, `w`arning, `e`rror. @jzwar wanted this to be a friendly hint, and I thought info in friendlier than warning. I'd personally go with debug, since it is just unset value - however, per default, we have it off, so it won't be so helpful. do you still believe in `w`arning?

jzwar: I like info. As someone might use it in a Jupyter notebook first, without reading the documentation, ... This is a perfect hint.

splinepy/microstructure/tiles/tile_base.py
j042: Yes, that would be good and I think we used to have it at the very beginning. I had some cases where abc made things a bit more complicated, so I decided to stay with simple inheritance

splinepy/microstructure/tiles/ellips_v_oid.py
jzwar: hey, stop judging me. This is 100% intentional :D If I could have () in the file name I would have done used them here

j042:
Todo: background color, in modification plot nr.1 is wrong

splinepy/helpme/fit.py
j042: Added debug log

splinepy/helpme/fit.py
```diff
@@ -52,30 +75,41 @@ def parameterize_line(fitting_points, n_fitting_points, centripetal):
...
+                select[i] = j
+                tmp.append(
+                    parameterize_line(
+                        reorganized_queries[tuple(select)].reshape(-1, dim),
+                        centripetal,
+                    )
+                )
+            u_ks.append(_np.mean(tmp, axis=0))
+
+        return u_ks
```
jzwar, line 95: This implementation is wrong, because the slices are set in the first loop to `[0,:]`, `[1,:]` etc...
This is also the reason the current tests fail.

splinepy/microstructure/microstructure.py
danielwolff1: okay, I wasn't aware of the Jupyter notebook issue, I guess it is fine then. Most important is that the user is informed at all :)

splinepy/microstructure/tiles/tile_base.py
danielwolff1: Okay, thanks for clarifying! 👍

j042:
@jzwar @clemens-fricke are you happy with the background?

j042:
todo: change extract bases' angle

j042:
todo: add `iges`

PR #423, opened 2024-04-26
finalize #394

j042:
# Overview
Mini PR that implements comments of #394.


jzwar:
Did you just delete the entire code? There is no LM implementation anymore🤔

j042:
No way. This is for #394

j042:
https://github.com/tataratat/splinepy/blob/82bcc382fd969c4ed5fda5ab839492b324cdc355/src/proximity/proximity.cpp#L338

Issue #433, opened 2024-05-27
xml io issues for CATS splines 


felipegonzalezco:
I found the following problems when generating a spline for CATS SplineLib:

1. the keyword `closed` expects at least `splDim` entries
2. we use `kntVec` instead of `knotVec` for each individual knot vector
3. `weights` are not optional and must be present

jzwar:
@felipegonzalezco Would it be possible to quickly post a minimal example of the current state and how it should be fixed, to better understand what is the issue, please? That would be very helpful.

felipegonzalezco:
Sure. As examples, here are screenshots of each case.
First, the keyword `closed` expects `splDim` entries. Currently, `splinepy` is writing only one value.
<img width="794" alt="closed" src="https://github.com/tataratat/splinepy/assets/53946371/9e4c2a1d-129e-4f93-a959-678c3c2b4d26">

Second, `kntVec` is used for individual knot vectors. It's kinda consistent with `kntVecs` which is the vector of knot vectors.
<img width="771" alt="kntVec" src="https://github.com/tataratat/splinepy/assets/53946371/a921219c-f5fb-4911-8f23-e7a766c16cde">

And third, I get an error if `weights` are not defined:
<img width="618" alt="weights" src="https://github.com/tataratat/splinepy/assets/53946371/d9d7bd3b-1f52-4fb7-b921-03e296f0054a">

I hope it is more clear now.

jzwar:
Just to make sure, for the second we need to rename `knotVec` to `kntVec`?

jzwar:
For the third, are weights only exported for rational splines or do they need to be exported for non-rational splines as well.

jzwar:
I updated the code to deal with the first two issues, please check out the branch bf-433

j042:
After looking through the sources of cats-splinelib, it seems that the static buffer - that fortran splinelib creates for xml.io - is too small to load a "long" line or expects less value per line. Not conclusive yet

jzwar:
@felipegonzalezco did you try it out?

j042:
fixed in #436 and #433. will be available in v0.1.1

felipegonzalezco:
Sorry for the late reply. I tried it out and it worked, so the issues are resolved. Thank you for fixing them.

PR #436, opened 2024-05-28
extension regarding #433

j042:
new line for each values


jzwar:
why though? 😅 

also: references need updating

j042:
After briefly looking at Fortran splinelib, they allocate a static buffer to read each line and pass around some pointers to load values, and it somehow has a problem with "a slightly longer line" of values. A bit of "it's not me, it's you" situation

PR #447, opened 2024-07-29
Add integration of physical function

markriegler:
# Overview
Integration of physical function

## Checklists
* [ ] Added example(s)
* [x] Added test(s)

## Summary by CodeRabbit

- **New Features**
	- Enhanced the integration functionality to accept spline objects, improving usability for spline integration.
	- Expanded documentation for the integration function, including detailed parameter descriptions and type safety checks.

- **Tests**
	- Introduced a new test case to validate the integration of a specific parabolic function over a defined domain, ensuring the accuracy of the integration process.


markriegler:
The issue might have something to do with the `splinepy.helpme.create.parametric_view` which I used for the integration. I just saw that the function returns a `BSpline` with degree 1 in every direction, which is different to the degrees of sphere (which has degrees [1 2 2]). I'll have to check

j042:
> Thanks for doing the test.
> 
> 
> 
> The issue might have something to do with the `splinepy.helpme.create.parametric_view` which I used for the integration. I just saw that the function returns a `BSpline` with degree 1 in every direction, which is different to the degrees of sphere (which has degrees [1 2 2]). I'll have to check

You should be able to pass an argument to get a conforming basis parametric view. But why would you need that?

clemens-fricke:
Thanks @jzwar. We integrated the fixes from #459 into this PR. The error shown above is still partly present, but it seems to be a numerical error. The relative error is at least e-5, and can be reduced by adding more knots into the spline.

PR #459, opened 2024-09-30
add para to parent scaling

j042:
# Overview
Integration of bsplines weren't correct - it was missing parametric space to "parent" (or integration) element scaling.
Added corresponding test
## Addressed issues
*  consider parametric space to parent element mapping

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

### DIFF
--- a/tests/helpme/test_integrate.py
+++ b/tests/helpme/test_integrate.py
@@ -178,7 +178,7 @@ def test_function_integration(np_rng):
     col1_factor = 2
 
     def volume_function(x):
-        vf = np.ones((x.shape[0], 2))
+        vf = np.ones((len(x), 2))
         # scale it with a factor to get a different value
         vf[:, 1] = col1_factor
         return vf
@@ -186,7 +186,15 @@ def test_function_integration(np_rng):
     bezier = splinepy.Bezier(
         degrees=[1, 2], control_points=np_rng.random((6, 2))
     )
+
     assert np.allclose(
         [bezier.integrate.volume(), col1_factor * bezier.integrate.volume()],
         bezier.integrate.parametric_function(volume_function),
+    )
+
+    # try bsplines
+    bspline = bezier.bspline
+    assert np.allclose(
+        [bspline.integrate.volume(), col1_factor * bspline.integrate.volume()],
+        bspline.integrate.parametric_function(volume_function),
     )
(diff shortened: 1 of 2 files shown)


j042:
I think this is also missing in iga-galerkin example

tests/helpme/test_integrate.py
```diff
@@ -178,15 +178,23 @@ def test_function_integration(np_rng):
     col1_factor = 2
 
     def volume_function(x):
-        vf = np.ones((x.shape[0], 2))
+        vf = np.ones((len(x), 2))
         # scale it with a factor to get a different value
         vf[:, 1] = col1_factor
```
clemens-fricke, line 183: Also a factor would assume a multiplication and not just the value. In this special case, the result is the same, but it might be a little weird for someone to read it and see that it just overwrites the value instead of multiplying it. So `col2_value` would be more accurate.

tests/helpme/test_integrate.py
```diff
@@ -178,15 +178,23 @@
...
         [bezier.integrate.volume(), col1_factor * bezier.integrate.volume()],
         bezier.integrate.parametric_function(volume_function),
     )
+
+    # try bsplines
+    bspline = bezier.bspline
+    assert np.allclose(
+        [bspline.integrate.volume(), col1_factor * bspline.integrate.volume()],
+        bspline.integrate.parametric_function(volume_function),
+    )
```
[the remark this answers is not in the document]
clemens-fricke: I am guessing that the integration of a function inside the parametric domain is spline agnostic. At least for the implemented spline? The original function only iterates through the bezier view of the spline? Or is it just not implemented for the other splines?

clemens-fricke (approved):
The proposed change is optional. If the question from the second comment is trivial, I have no problems with this PR. Thanks for the work. Looks good.

tests/helpme/test_integrate.py
j042: I thought col starts at 0?

clemens-fricke:
@markriegler did you check this out? This might be relevant for your integration PR #447 as well.

markriegler:
> @markriegler did you check this out? This might be relevant for your integration PR #447 as well.

I just realized that I did the same parent to para scaling in my PR #447 , but in another unnecessarily more complicated way.

PR #467, opened 2025-03-21
Fix query check lower bound

mkofler96:
# Overview
There is a small mistake that shows the wrong bounds when the query is out of bounds.

## Addressed issues
*  Issues addressed

## Showcase
```Python
import splinepy

bez = splinepy.Bezier([0], [[0]])
# works
bez.evaluate([[0]])
# gives value error
bez.evaluate([[-1]])
```
gives 

```Python
ValueError: Query request out of bounds in parametric dimension 0. Detected query [-1.] at positions 0, which is out of bounds with minimum values [1.].
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated error messages to correctly display the minimum parameter limits. Users will now receive clearer and more accurate feedback when providing inputs that fall outside the accepted range.

### DIFF
--- a/splinepy/helpme/check.py
+++ b/splinepy/helpme/check.py
@@ -34,7 +34,7 @@ def valid_queries(spline, queries):
     # Check minimum value
     min_query = _np.min(queries, axis=0)
     if _np.any(bounds[0, :] > min_query):
-        error_dim = _np.where(bounds[1, :] > min_query)[0][0]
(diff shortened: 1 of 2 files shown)


clemens-fricke (approved):
Thanks for the find. Am I correct that this is not a functional error but only an error in the error message?

clemens-fricke:
Thanks for fixing the offline discussed issue.

PR #476, opened 2025-10-08
Allow any negative number in the knot-vector

### LINKED ISSUE #475: Knot Vectors with knots less than -1 are not supported
I messed around with splinepy once again... and found this weird error message:
```
import splinepy
print(splinepy.__version__)

s = splinepy.BSpline(degrees=[2], knot_vectors=[[-2, -1, 0, 1, 2, 3]], control_points=[[0.0, 0.0], [0.5, 0.0], [1.0, 0.0]])
```

output:
```
0.2.0
RuntimeError: SPLINEPY ERROR - -1 -2 Knots of parametric dimension ( 0 ) are not in increasing order.
```

I suspect the issue is, that the check assumes the initial knot to be -1 and not -inf:
https://github.com/tataratat/splinepy/blob/99ae5de064e89e821cbdb9b7d308d41c45e5ddbf/src/py/py_spline.cpp#L109

~~I'm not sure if this has any relevance for real world problems though...~~
One example where this useful would be the example section in the NURBS Book, Chapter 12.2. Here, uniform unclamped knot vectors are defined, which start at -3.

reox:
This resolves #475

# Overview
Description

## Addressed issues
*  Issues addressed

## Showcase

Knot-Vectors with values less than -1 are now possible
```
import splinepy

s = splinepy.BSpline(degrees=[2], knot_vectors=[[-2, -1, 0, 1, 2, 3]], control_points=[[0.0, 0.0], [0.5, 0.0], [1.0, 0.0]])
```

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

## Summary by CodeRabbit

* **Bug Fixes**
  * Relaxed initial knot-order validation to reduce false ordering errors with extreme or tightly spaced knot values, improving spline creation robustness and stability.

* **Tests**
  * Added a constructor test covering multi-degree B-spline construction with unclamped, uniform-like knot vectors to increase coverage for edge-case spline setups.

### DIFF
--- a/tests/test_spline_constructors.py
+++ b/tests/test_spline_constructors.py
@@ -17,7 +17,10 @@ def test_BSplines(np_rng):
     # No assertion query; unclamped, uniform knot-vector
     splinepy.BSpline(
         degrees=[2, 2],
-        knot_vectors=[[-4, -3, -2, -1, 0, 1, 2, 3, 4], [-15, -10, -5, 0, 5, 10, 15]],
+        knot_vectors=[
+            [-4, -3, -2, -1, 0, 1, 2, 3, 4],
+            [-15, -10, -5, 0, 5, 10, 15],
+        ],
         control_points=np_rng.random((6 * 4, 2)),
     )
     # Check control points


src/py/py_spline.cpp
```diff
@@ -106,7 +107,7 @@ void PySpline::NewCore(const py::kwargs& kwargs) {
       knot_vector.reserve(n_knots);
 
       int nknots{0};
-      prev_knot = -1.;
+      prev_knot = std::numeric_limits<double>::min();
```
reox, line 110: Actually, one should use lowest() here, right? Otherwise a knot vector that contains min() would not work...

Issue #478, opened 2025-10-27
Visualization of Arrow Data not working in 3D


kkILSB:
The following code works as expected (arrows for field data) for `number_of_dimensions=2`, but not for `number_of_dimensions=3`.

```
from splinepy import BSpline
from splinepy.helpme.create import box as CreateBox
from splinepy.microstructure import Microstructure as Lattice
from splinepy.microstructure.tiles import Cross2D as Cross2d, Cross3D as Cross3d

number_of_dimensions = 3
parameter_spline = BSpline(degrees=[1] * number_of_dimensions, knot_vectors=[[0.0, 0.0, 1.0, 1.0]] * number_of_dimensions, control_points=[[0.2]] * 2**number_of_dimensions)
def EvaluateParameterSplineSensitivities(parametric_coordinates):
    from numpy import put_along_axis as PutAlongAxis, zeros as CreateZeros

    number_of_parametric_coordinates = parametric_coordinates.shape[0]
    basis_function_matrix = CreateZeros((number_of_parametric_coordinates, len(parameter_spline.control_points)))
    basis_functions, support = parameter_spline.basis_and_support(parametric_coordinates)
    PutAlongAxis(basis_function_matrix, support, basis_functions, axis=1)
    return basis_function_matrix.reshape(number_of_parametric_coordinates, 1, -1)

deformation_function = CreateBox(*[1.0 for _ in range(number_of_dimensions)])
lattice = Lattice(deformation_function=deformation_function, tiling=[1] * number_of_dimensions, microtile=Cross2d() if number_of_dimensions==2 else Cross3d(), parametrization_function=lambda parametric_coordinates : parameter_spline.evaluate(parametric_coordinates))
lattice.parameter_sensitivity_function = lambda parametric_coordinates : EvaluateParameterSplineSensitivities(parametric_coordinates)
composed = lattice.create()
composed.show(resolutions=10)
composed.show_options["arrow_data"] = "field"
composed.spline_data["field"] = 1
composed.show()
```

Jacques made the mistake of coming to the institute again and even offered to investigate 😅

EDIT:
Upon further investigation, I found that
```
from splinepy.helpme.create import box as CreateBox

number_of_dimensions = 3

box = CreateBox(*[1.0] * number_of_dimensions)
box.show()
box.spline_data["data"] = box
box.show_options["arrow_data"] = "data"
box.show()
```
seems to work as expected. In contrast,
```
from splinepy import SplineDataAdaptor
[...]

PR #482, opened 2025-11-03
Retain interfaces when adding fields

jzwar:
# Overview
When adding fields, the interfaces the list of patches where saved as an _empty_ Multipatch, meaning, they had no information about the connectivity of their underlying geometry. 

## Addressed issues
- #478 

## Checklists
* [ ] Added example(s)
* [ ] Added test(s)

## Summary by CodeRabbit

* **New Features**
  * Added a Python-accessible API to manually set interface orientations.

* **Bug Fixes**
  * Clearer, stricter validation of interface dimensionality with improved mismatch errors.
  * Ensured newly created fields inherit existing interface data so interface settings propagate consistently.
  * Returned cached interface orientations when available to avoid redundant recomputation.

* **Documentation**
  * Updated PR template with a formatted showcase example and minor formatting improvements.

### DIFF
--- a/src/py/py_multipatch.cpp
+++ b/src/py/py_multipatch.cpp
@@ -602,24 +602,21 @@ void GetInterfaceOrientation(
   const int& dim_ = pyspline_start->SplinepyDim();
 
   // Checks
-  if ((para_dim_ < 2) || (dim_ < 2)) {
+  if (para_dim_ < 2) {
     splinepy::utils::PrintAndThrowError(
-        "Spline Orientation can not be checked, either spline is "
-        "one-dimensional ",
-        para_dim_,
-        "D -> ",
-        dim_,
-        "D, the adjacent one has dimensions ",
-        pyspline_end->SplinepyParaDim(),
-        "D -> ",
-        pyspline_end->SplinepyDim(),
-        "D.");
+        "Curves are not supported for automatic computatation of interface "
+        "orientations");
+  }
+  if (dim_ < 2) {
+    splinepy::utils::PrintAndThrowError(
+        "Scalar splines are not supported for automatic computation of "
+        "interface orientations");
   }
   if ((para_dim_ != pyspline_end->SplinepyParaDim())
       || (dim_ != pyspline_end->SplinepyDim())) {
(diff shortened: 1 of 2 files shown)


clemens-fricke (approved):
Looks good to my eyes, though I am not a subject matter expert so can only give a superficial OK. Specially since I have never used pybind so the cpp-py interface is something I have no knowledge about

src/py/py_multipatch.cpp
```diff
@@ -602,11 +602,10 @@ void GetInterfaceOrientation(
   const int& dim_ = pyspline_start->SplinepyDim();
 
   // Checks
-  if ((para_dim_ != pyspline_end->SplinepyParaDim())
-      || (dim_ != pyspline_end->SplinepyDim())) {
+  if ((para_dim_ < 2) || (dim_ < 2)) {
     splinepy::utils::PrintAndThrowError(
-        "Spline Orientation can not be checked, as they have mismatching"
-        "dimensionality start spline has dimensions ",
+        "Spline Orientation can not be checked, either spline is "
```
kkILSB: What should the actual error message look like? I think this currently does not form a proper statement. And if `para_dim > 1` or `dim > 1` are also possible, maybe "one-dimensional" could be replaced by something like "curve" (`para_dim` is 1) or "scalar" (`dim` is 1) 🤔

src/py/py_multipatch.cpp
```diff
@@ -696,6 +705,42 @@ void GetInterfaceOrientation(
...
+    splinepy::utils::PrintAndThrowError(
+        "Interfaces need to be computed or set first, because they will be "
+        "reset upon recomputation");
+  }
+
+  // Check if the shape of the interface orientation is usable
+  int para_dim = ParaDim();
+  const int n_faces_per_element = 2 * para_dim;
+  // 4 includes - patch id start, face id start, patch id end, face id end
+  const int n_entries_per_line = 4 + 2 * para_dim;
```
kkILSB: Use `n_faces_per_element` for the second summand?

jzwar: 🤦 ups

